PR #37, opened 2026-07-23
Replace WorkOS with GitHub and per-instance OAuth

javiertoledo:
## Summary

- replace WorkOS and the development login bypass with direct GitHub App user OAuth or an instance-bound OIDC broker mode
- add exact GitHub identity, account, and installation admission plus a transactional `facility instance bootstrap` command
- make every instance its own OAuth authorization server for MCP, with PKCE, dynamic client registration, audience-bound ES256 JWT access tokens, encrypted adapter storage, refresh rotation/reuse detection, consent, discovery, and revocation
- deploy the MCP service in Docker Compose and AWS Terraform, and document local, self-hosted, and SaaS broker configuration
- record the critical-change unit/integration regression policy in `AGENTS.md` and `CLAUDE.md`

## Breaking changes

This intentionally removes WorkOS configuration, schemas, and runtime behavior. Existing instances must configure GitHub/OIDC variables, apply migration `0025_github_identity_and_oauth.sql`, seed without demo data, and bootstrap the dedicated GitHub binding. There is no compatibility or transition layer because the system is not in production.

The commercial GitHub-to-OIDC broker remains outside this repository. This PR implements and validates the instance-side broker contract.

## Security and regression coverage

- fake GitHub token, user, verified-email, paginated-installation, and callback flow
- signed fake OIDC broker with nonce and instance mismatch denial
- invitation, ambiguous identity, exact account/installation, cross-tenant, malformed, expired, wrong-audience, wrong-scope, and foreign-signature denials
- complete MCP OAuth authorization-code flow with PKCE and consent, public JWKS, forwarded-host poisoning defense, encrypted token persistence, refresh rotation, and replay rejection
- transactional/idempotent bootstrap integration against PostgreSQL
- login audit event regression and no upstream token persistence

## Verification

- `pnpm verify`
- `terraform -chdir=infra/terraform/aws validate`
- `SECRET_MASTER_KEY=... docker compose config --quiet`
- 210 API tests passed before the final repository-wide verification; all package suites and repository guards pass in `pnpm verify`
- `pnpm audit`: no known vulnerabilities (includes Next.js 16.2.11 security patch)


javiertoledo:
Field-tested the whole flow end to end on a local instance (fresh GitHub App, real OAuth round-trip through an HTTPS tunnel, sign-in, app usage, sign-out). Findings, all addressed on the branch except the first:

1. **Migration 0028 backfills `github_installations.account_id` with `installation_id` as a placeholder** — a value that can never match the real account id, so every legacy installation fails admission (`installation_access_required`, surfaced to the user as the generic `auth_failed`) until the next `installation` webhook heals the row (the processor does upsert the real account id). Worth either resolving real ids in the migration, or documenting that pre-existing installs need one webhook delivery (or a manual fix) before first login.
2. **Sign-in silently requires the App's `Email addresses: Read-only` account permission** — without it `/user/emails` 403s inside the exchange and the user gets the same generic `auth_failed`. Now documented (README permission table + troubleshooting); a distinct error for this branch of the exchange failure would save future operators the spelunking.
3. **Sign-out was broken**: the topbar posted an HTML form (urlencoded) at the JSON-only API → 415 before the handler. Fixed in `c1ac704` (client action: bodyless fetch POST + redirect to /login).
4. **Dev via HTTPS tunnel needs `allowedDevOrigins`** (Next blocks cross-origin dev requests — HMR + RSC fetches, which broke all client-side navigation). Fixed in `c1ac704` for the tunnel hostname + documented in the README troubleshooting table.
5. **README** (`63a8c26`): the self-host quick start is now the complete install path — App creation with the full permission table, client secret + private key steps, every `.env` value, `facility instance bootstrap` with how to find each id, and a troubleshooting table mapping the real error codes to fixes.

With those in, the flow works cleanly: authorize → callback → session → app, and logout round-trips to /login.

PR #44, opened 2026-07-27
docs: one home for everything about using Facility

javiertoledo:
**Stacked on #43** — review that one first; this PR targets its branch so the diff stays honest.

The repository carries two documentation trees. The Docusaurus site under `apps/docs` has 23 pages; a second `docs/` tree at the root has nine more — ~1,130 lines, including the platform architecture and the interface contracts — that the site never included. Publishing the site as it stands would ship documentation with its architecture missing, which is the reason this had to happen before the Pages workflow does anything.

## The rule

`docs/` at the root is for documents about **building** Facility. Everything about **using** it belongs on the site. `docs/README.md` now states it, so the split does not silently drift back.

## Moved onto the site

| from | to | why there |
|---|---|---|
| `docs/method.md` | `concepts/method.md` | first concept in the sidebar — it is the spine the rest hangs from |
| `docs/faq.md` | `faq.md` | top level, next to the roadmap |
| `docs/guards.md` | `reference/guards.md` | |
| `docs/hardening.md` | `reference/hardening.md` | |
| `docs/platform/ARCHITECTURE.md` | `reference/architecture.md` | |

## Merged, not moved

**The watchtower** existed twice: the root document described the repository lane, the site page the platform lane. The site page now carries both. It gains the canary's authorization design — *the message is authorized, not the sender*, with the byte-identical hash pinning that means a leaked canary key can at worst replay one fixed read-only probe — and the repository-lane specifics: the guard that locks the watchtower, and budgets as a reviewed file rather than a dashboard setting.

Dropped from it: the old "what this is not (yet)" section, which described per-run receipts as future work. They exist.

**INTERFACES.md** was largely superseded by the site's `api`, `cli` and `mcp` pages — I checked each claim rather than assuming, and the MCP deployment variables, the SDK example and its caveats are all already there. Two pieces were genuinely missing and are preserved:

- the integration webhook wire contract, including the `facility.signal.v1` lifecycle envelope, as `reference/webhooks.md`;
- the troubleshooting list, as a section of `reference/api.md`.

## Stayed at the root

`docs/testing.md` (what verification must prove before a change is releasable) and `docs/platform/PRD.md` (the v1 requirements). Both are about building the thing.

The PRD is worth a decision at some point: it is a v1 draft that the site now supersedes, and its "Vercel for the AI software factory" framing is not how we describe Facility today. Keeping it as a record is defensible; so is deleting it. Not something to decide inside a docs move.

## Verification

- The site builds with `onBrokenLinks: "throw"`, and all seven new or moved routes render — including the merged canary section.
- Every relative link in `README.md`, `CONTRIBUTING.md` and `SECURITY.md` resolves to a file that exists (checked by script, not by eye). Five README links, one each in SECURITY and CONTRIBUTING, and one comment in the `facility-crew.yml` template were repointed.
- `pnpm lint`, `pnpm guards`, `pnpm --filter @facility/docs typecheck`, and the CLI's 64 tests — the template comment change touches rendered output.


javiertoledo:
Fixed in 769455b — the PRD now links to `apps/docs/docs/concepts/method.md`. Commit 5847f30 also adds a tracked-Markdown link guard so this broken-path regression fails locally and in CI.

PR #53, opened 2026-07-30
ci: let the commits decide the version

javiertoledo:
Today a release means editing two `package.json` files by hand, remembering they must agree, tagging, and writing notes from memory. Fine for one maintainer and no users; wrong for an open-source project where a contributor's merged fix should reach a release without anyone reconstructing what changed.

## How it works

`release-please` keeps a release pull request open on `main`. It accumulates the conventional-commit subjects merged since the last release, writes `CHANGELOG.md`, and bumps the version in `package.json` **and** `packages/cli/package.json` — the two that `scripts/release.mjs` refuses to release unless they agree, which is exactly the manual step most likely to be forgotten.

**Merging that pull request is the release decision.** It creates the `vX.Y.Z` tag, and the tag starts the pipeline that already exists: verification, self-host build, sandbox E2E, then npm and — once the repository is public — the images. Nothing new publishes, nothing publishes on its own, and the human gate stays exactly where it was.

## Choices worth reviewing

**Pre-1.0 semantics.** `bump-minor-pre-major` is on, so `feat!:` is a minor bump rather than a major one. That is the plain reading of what the README already tells users: no upgrade path is promised between `0.x` releases. `fix` and `feat` are both patch until 1.0.

**Pull request titles are checked.** A squash merge makes the title the subject on `main`, and the changelog is built from subjects — so the title is the only place to catch it before it lands. The check is fifteen lines of `grep -E`, no external action. Tested against titles this repository actually uses: `fix(gateway): drain metering`, `chore(deps)!: drop node 20` and `revert: feat: x` pass; `Update README` and `wip` do not.

Worth pairing with a settings change: the repository currently allows merge commits, and `Merge pull request #51 from …` is not a subject anyone wants in a changelog. Squash-only would make the checked title the thing that lands. Your call — the automation works either way, since release-please reads the branch commits and skips merge commits.

**Why not semantic-release**: it publishes directly from CI on merge, which would duplicate the release contract `scripts/release.mjs` already enforces and requires write access to `main` from a workflow. Why not changesets: it fits many independently-versioned packages; this repository has one version, and the per-pull-request changeset file is a ritual an outside contributor would not expect.

## What it needs from you

Its own credential, as `RELEASE_PLEASE_TOKEN` with `contents: write` and `pull-requests: write`. The default `GITHUB_TOKEN` cannot open pull requests here because *Allow GitHub Actions to create and approve pull requests* is deliberately off — and that switch is one toggle covering both verbs, so re-enabling it to let a bot open a release PR would also re-open the door to automated approvals. A fine-grained PAT works; a small GitHub App is better, since it is not tied to a person.

The workflow fails with that explanation rather than a bare 403 when the secret is missing.


javiertoledo:
Closing in favour of the simpler shape Javier chose: a release on every merge to `main`, with the commit subjects deciding whether it is a patch or a minor, and no release pull request in between.

Two things from here survive and arrive separately:

- the conventional-commit check on pull request titles, which either approach needs — the subjects are what the version is computed from;
- an explicit instruction in `AGENTS.md` and `CLAUDE.md`, so an agent writing a commit knows the subject is not decoration.

The design note worth keeping out of the closed diff: a tag pushed with the default `GITHUB_TOKEN` does not trigger other workflows, so "CI pushes a tag and the tag pipeline picks it up" silently does nothing. The replacement therefore decides, verifies, publishes and writes the tag in one run, and the tag becomes the record of what was published rather than the request to publish it.

PR #83, opened 2026-08-06
perf(runner): persist lifecycle phase timings

adrian-lorenzo:
## What changes

- Persist nine monotonic runner lifecycle timings as the existing `run_events` type `phase`, without a database migration.
- Keep event payloads fixed and safe: phase name, terminal status, duration, controlled outcome, and typed skip reason only.
- Keep timing writes best-effort so transient observability failures cannot replace successful work or the operation's real error.
- Map phase events explicitly in the run cockpit so a failed acceptance outcome does not masquerade as a failed check event.
- Cover the recorder, authenticated API delivery, persistence, cockpit presentation, and real sandbox lifecycle order.

## Why

Facility needs a trustworthy baseline before optimizing package installation, provisioning, agent execution, result capture, acceptance, and delivery. These timings expose where each run spends time while adding at most nine authenticated event writes and no new service or schema.

## Verification

- [x] `COMPOSE_ENV_FILES=.env.example pnpm verify` passes locally (Biome 344 files; API 306 tests; gateway 37; runner 115; web 26; guards 2; no known vulnerabilities).
- [x] Follow-up robustness coverage passes: runner 116 tests plus runner typecheck after making timing ingestion best-effort.
- [x] Behaviour verified beyond the test suite: built the real runner Docker image, then ran `FACILITY_E2E_DOCKER=1 pnpm test:e2e-sandbox` against isolated Postgres; the 33.99 s run persisted all nine phase events in order with non-negative durations.
- [x] Documentation updated in `runner/README.md`.


adrian-lorenzo:

PR #84, opened 2026-08-06
perf(runner): add adaptive CodeBuild fast path

adrian-lorenzo:
## What changes

Sandbox profiles can now explicitly disable nested Docker on the AWS CodeBuild driver with `setup.nested_docker: false`. The API validates the capability, the orchestrator passes it only to AWS, and the wrapper skips the nested daemon, restricted proxy, bind broker, and shared-mount setup while retaining the untrusted UID, metadata isolation, credential stripping, and writable workspace boundary. Docker-backed profiles now prefer overlay2, then fuse-overlayfs, then degraded vfs. A fixed allowlisted `sandbox_runtime` event records the selected boundary.

This is a stacked draft on #83 so its diff is one commit. It must not merge until #83 merges, this branch is rebased and retargeted to `main`, and fresh CI passes.

## Why

Lightweight agents currently pay the full nested-Docker startup cost even when they never invoke Docker. This creates an explicit subtractive capability instead of inferring behavior from agent names, preserves legacy profiles by default, and keeps the AWS provider boundary swappable.

## Verification

- [x] `COMPOSE_ENV_FILES=.env.example pnpm verify` passes locally: 315 API tests, 37 gateway tests, 122 runner tests, all remaining suites, guards, and zero known dependency vulnerabilities.
- [x] Behaviour verified beyond the test suite: an exact runner image passed the real privileged no-Docker boundary in under one second and the Docker-backed create-only security boundary in about ten seconds, including metadata, credentials, UID, bind, privilege-denial, and rollback checks.
- [x] Documentation updated in `runner/README.md`.

Native GitHub CI must pass both the full Docker sandbox loop and the no-Docker fast path before this draft can become ready. Claude Opus high completed an equal security/architecture review and returned APPROVE WITH CONDITION on that native-CI evidence, with no blocking findings.


adrian-lorenzo:

PR #90, opened 2026-08-06
fix(api): align database clock boundaries

adrian-lorenzo:
## What changes\n\nUse one clock domain for database-backed time boundaries:\n\n- initial durable PR deliveries now receive an application-clock next_attempt_at, matching the delivery drain and retry scheduler;\n- implicit spend windows now use PostgreSQL current_timestamp, matching database-managed created_at;\n- explicit spend bounds retain canonical ISO normalization and exact 720-hour default-window semantics;\n- deterministic regressions cover sub-millisecond delivery eligibility and a skewed application/database clock.\n\n## Why\n\nPostgreSQL timestamps preserve microseconds while JavaScript Date preserves milliseconds. Immediately after an insert, the database default could be a fraction of a millisecond later than the application's comparison value. This intermittently made a fresh delivery ineligible and hid a fresh LLM request from the default spend window.\n\nThe full verifier exposed 13 failures from the same boundary; this patch removes the race rather than retrying the suite.\n\n## Verification\n\n- [x] pnpm verify passes locally\n- [x] Behaviour verified beyond the test suite: a fresh-database focused run passed 133/133 affected API tests; the spend regression skews the application clock one second behind PostgreSQL without freezing timers, and the delivery regression claims at the exact stored millisecond boundary\n- [x] No documentation change: this restores the documented immediate-delivery and spend behavior without changing the API contract\n\nFull verifier evidence: 350 API, 37 gateway, 122 runner, 91 repository-script tests, all remaining packages, 2/2 guards, and no known dependency vulnerabilities. Claude Opus high independently reviewed the implementation and approved it.

### DIFF
--- a/services/api/src/routes/v1/providers-budgets-spend.ts
+++ b/services/api/src/routes/v1/providers-budgets-spend.ts
@@ -499,10 +499,14 @@ export async function registerProvidersBudgetsSpendRoutes(
         to?: string;
         groupBy?: "model" | "agent" | "task" | "day";
       } & PageQueryValue;
-      const toDate = q.to ? new Date(q.to) : new Date();
-      const fromDate = q.from ? new Date(q.from) : new Date(toDate.getTime() - 30 * 86_400_000);
-      const from = fromDate.toISOString();
-      const to = toDate.toISOString();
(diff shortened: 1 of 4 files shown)


adrian-lorenzo:

PR #92, opened 2026-08-06
perf(db): unify the deployment gate

adrian-lorenzo:
## Summary
- replace separate production migrate and seed commands with one private database deployment entrypoint
- keep the existing Compose migrate service and AWS one-shot ECS task; no service or component is added
- hold one bounded PostgreSQL advisory lock across migration validation, migration application, and idempotent system-data reconciliation
- add immutable SHA-256 migration checksums with safe legacy backfill and fail-closed mismatch handling
- emit structured phase timing and outcome events with distinct operational exit codes

## Why
The existing migrate task ran two independent Node processes. That meant two database connections, two coordination boundaries, and no protection against concurrent migration plus seed reconciliation. The new private package entrypoint makes database deployment one atomic operational gate while preserving the current topology.

This is intentionally not exposed as a public Facility CLI command: it is an image-internal deployment primitive that requires direct database credentials. The public CLI should not normalize operator database access.

A global seed fingerprint was considered and rejected. Registry reconciliation is tenant-aware and repairs drift; skipping it globally could omit new tenants or changed runner and driver data. Existing content hashes already avoid unchanged writes, and measured warm reconciliation is 209 ms.

## Operational contract
- default lock timeout: 60 seconds
- configurable timeout ceiling: 15 minutes
- exit 2: invalid configuration
- exit 10: deployment lock timeout
- exit 11: migration checksum mismatch
- exit 12: migration execution failure

## Validation
- full repository verifier: passed
- database tests: 14 of 14 passed
- API tests: 350 of 350 passed
- gateway tests: 37 of 37 passed
- critical CLI and security tests: 67 of 67 passed
- repository script tests: 91 of 91 passed
- runner tests: 122 of 122 passed
- repository guards: 2 of 2 passed
- all-severity dependency audit: no known vulnerabilities
- built deployment smoke test: 242 ms total; lock 28 ms; migrations 5 ms; reconciliation 209 ms
- missing database configuration: structured failure with exit 2
- Codex review: approved
- Claude Opus high-reasoning design and implementation review: approved

PR #93, opened 2026-08-06
perf(sandbox): cache project dependencies

adrian-lorenzo:
## Summary

- add one mutable CodeBuild S3 dependency cache per Facility project using an HMAC-derived, unguessable location prefix
- keep the CodeBuild project default at NO_CACHE so omitted run overrides fail closed instead of sharing tenant data
- cache only pnpm content-addressed storage and npm _cacache; exclude workspaces, credentials, browsers, Supabase, and Docker state
- grant the CodeBuild role prefix-scoped S3 and S3-via-KMS permissions, then expire active caches after 30 days and old versions after 7
- make Facility doctor reject missing cache configuration or an unsafe project-level shared cache
- preserve status, logs, and stop operations for existing builds during partial upgrades

## Why

The field E2E run reported that CodeBuild had no cache paths. AWS local CodeBuild caching is unavailable for VPC builds, so this uses the existing S3 bucket and CodeBuild driver rather than adding a cache service, bucket, database state, or image-builder service.

AWS reference:
- https://docs.aws.amazon.com/codebuild/latest/userguide/caching-local.html
- https://docs.aws.amazon.com/codebuild/latest/userguide/caching-s3.html
- https://docs.aws.amazon.com/codebuild/latest/APIReference/API_ProjectCache.html

## Isolation

- the cache partition is HMAC-SHA256 over the organization and project using the existing Facility master key
- the partition is a trusted LaunchSpec field and never enters the sandbox environment
- every partition is anchored in a separate documented S3 location prefix
- malformed or missing partitions fail before StartBuild
- the CodeBuild role cannot list cache keys and cannot access Facility object prefixes outside codebuild-cache
- untrusted runner identities continue to have AWS credential variables removed and link-local metadata blocked

## Verification

- repository verifier passed from the exact commit:
  - lint
  - all 13 package typechecks
  - clean uncached production build
  - DB: 14 tests
  - API: 356 tests
  - gateway: 37 tests
  - critical CLI and security: 67 tests
  - repository scripts: 91 tests
  - runner: 123 tests
  - all remaining packages
  - repository guards: 2 of 2
  - all-severity dependency audit: clean
- Terraform validate passed with the pinned AWS provider
- focused cache suite passed 57 API tests, including database-backed sandbox integration
- infrastructure guards passed 5 of 5
- Claude Opus high independently reviewed the proposal, threat model, implementation, and final simplifications; verdict APPROVE

## Stack

- base: #92
- this PR contains one Conventional Commit with patch release impact
- internal planning documents are intentionally excluded


adrian-lorenzo:

PR #105, opened 2026-08-07
feat(infra): make storage and web routing portable

adrian-lorenzo:
## Summary

- start the bundled MinIO and bucket initializer only for local S3-compatible storage
- let external S3-compatible deployments omit both storage containers without weakening the local readiness gate
- read the web control-plane origin at container runtime so one immutable image can move between deployments
- preserve same-origin browser sessions through a hardened streaming Route Handler proxy
- remove the obsolete build argument from Compose, Bake, the AWS build fallback, and image publication

## Security and portability boundaries

- proxy destinations are fixed to a configured bare HTTP(S) origin; credentials, paths, queries, fragments, forwarding headers, hop-by-hop headers, and redirect following are rejected or stripped
- production fails closed when the API origin is missing or invalid
- request bodies, response bodies, session cookies, status, redirects, and multiple `Set-Cookie` values are preserved
- local MinIO still blocks API/gateway startup until bucket creation succeeds when its profile is active
- the internal Facility planning/review documents are deliberately excluded

## Evidence

- `env COMPOSE_DISABLE_ENV_FILE=true COMPOSE_ENV_FILES=.env.example corepack pnpm@11.1.2 verify`
  - 372 API tests, 37 gateway tests, 33 web tests, 124 runner tests
  - repository guards: 2 passed
  - dependency audit: no known vulnerabilities
- built-container runtime routing: POST returned upstream status 209, preserved cookie/body/`Set-Cookie`, and removed `/api` from the upstream path
- built-container SSE routing: first event visible at 75 ms; second event visible at 1,574 ms after the upstream delay
- Compose profile contract: 8 default services; 10 with `local-storage`; external S3 starts no MinIO services
- independently reviewed with Claude Opus at high reasoning; no P0/P1 findings

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -82,30 +82,22 @@ jobs:
             dockerfile: Dockerfile
             target: api
             context: .
-            build_args: ""
           - image: gateway
             dockerfile: Dockerfile
             target: gateway
             context: .
-            build_args: ""
           - image: mcp
             dockerfile: Dockerfile
             target: mcp
             context: .
-            build_args: ""
           - image: web
             dockerfile: apps/web/Dockerfile
             target: web
             context: .
-            # The official image serves Facility's hosted control plane. AWS
-            # self-hosters with a different api_url must build their web image
-            # for that deployment, as documented in the runbook.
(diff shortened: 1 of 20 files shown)


adrian-lorenzo:

PR #148, opened 2026-08-14
fix(gateway): meter Anthropic streaming input tokens from message_start

### LINKED ISSUE #54: Meter Anthropic streaming input tokens from message_start
## Context

Raised while reviewing #52, when its shutdown regression exercised a streamed Anthropic response with both input and output usage.

## Problem

Anthropic reports streaming input usage under `message_start.message.usage`, while `UsageTee` currently reads only top-level `usage` and `delta.usage`. The input tokens are therefore omitted from metering. In the existing 1,000,000-input / 1,000,000-output fixture, the gateway records 1,500 cents instead of the expected 1,800 cents, undercounting both the request row and budget spend.

## Proposed change

Merge `message.usage` into Anthropic SSE usage parsing, preserving the existing direct and delta sources.

## Why not in #52

This is a pre-existing token-parser defect. #52 is scoped to making already-computed metering durable across streamed-response completion and shutdown.

## Acceptance criteria

- [ ] Anthropic `message_start.message.usage.input_tokens` and cache-token fields are included in streamed usage and cost.
- [ ] Later `message_delta.delta.usage` output fields still merge without erasing input fields.
- [ ] Unit coverage includes valid, split-chunk, and malformed SSE frames.
- [ ] A deterministic local-provider/Postgres integration proves the request row and spend counter include both input and output cost, with no live credentials or network access.

haiderysfzi:
## What changes

- Merge `message_start.message.usage` into the Anthropic streaming usage sources, so input and cache tokens are metered alongside the output tokens `message_delta` already reported.
- Merge the three usage sources separately instead of spreading them into one object, so a later output-only frame cannot erase input tokens an earlier frame reported.
- Assert input tokens, request-row cost, and the spend counter in the existing streaming integration test, and add unit coverage for valid, split-chunk, malformed, and non-streamed frames.

## Why

Closes #54.

Anthropic reports streamed input and cache usage under `message_start.message.usage`, while `UsageTee` read only the top-level `usage` and `delta.usage`. Streamed input tokens never reached metering, so with the existing 1M-input/1M-output fixture the gateway recorded 1500 cents against the 1800 cents owed. That undercounts the `llm_requests` row and the `spend_counters` row a hard budget enforces against, so a project can outspend its cap on streamed traffic.

The gap survived because the streaming test asserted only `outputTokens`. It now asserts the input tokens, the row cost, and the counter, so an undercount fails the suite rather than the invoice.

## Verification

- [ ] `pnpm verify` passes locally — it reached the critical integration step and stopped on API timeout flakes unrelated to this change, detailed below.
- [x] Behaviour verified beyond the test suite — reverted `usage.ts` alone and re-ran the new unit file: 4 of 5 tests fail with input tokens at 0, and the single pass is the non-streamed top-level `usage` case the change does not touch. Restored, all pass. The integration assertion moves 1500 → 1800 cents on both the request row and the spend counter.
- [x] Documentation updated, or no user-facing change — no user-facing surface changes; the correction is to recorded usage.

Commands run:

- `pnpm --filter @facility/gateway test` — 57/57 passed against an isolated `facility_gw`
- `pnpm --filter @facility/gateway typecheck`
- `pnpm lint`
- `pnpm guards` — 2 guards ran, 0 failed
- `git diff --check`

On the `pnpm verify` box: the failures are in `services/api`, which does not depend on `@facility/gateway` and cannot import the changed file. The gateway suite also runs after the API suite, against a separate database. Running the same 42-file API suite three times gave three different results: `test/assistant.test.ts` failed during `pnpm verify`, unmodified `main` passed 42/42, and this branch produced two 5-second timeouts in `test/watchtower.test.ts`. That is the same class of unrelated 5-second API timeout reported in #143. Happy to re-run anything on request.

Acceptance criteria from #54: `message_start` input and cache fields are metered; a later `message_delta` merges output without erasing them; unit coverage spans valid, split-chunk, and malformed frames; and the integration proof runs against local Postgres and a stub provider, with no live credentials or network.

### DIFF
--- a/services/gateway/src/usage.ts
+++ b/services/gateway/src/usage.ts
@@ -111,9 +111,16 @@ export class UsageTee extends Transform {
     try {
       const parsed = JSON.parse(text);
       if (this.provider === "anthropic") {
-        const direct = usageFromJson("anthropic", parsed);
-        const delta = usageFromJson("anthropic", { usage: parsed.delta?.usage });
-        mergeUsage(this.usage, { ...direct, ...delta });
+        // Anthropic splits streamed usage across frames: `message_start` carries
+        // input and cache tokens under `message.usage`, `message_delta` carries
+        // output tokens under `delta.usage`, and a non-streamed body reports
+        // everything under a top-level `usage`. Merge each source separately
+        // rather than spreading them into one object: a spread copies absent
+        // keys as `undefined` and would drop an earlier frame's input tokens
+        // when the later frame reports only output.
+        for (const source of [parsed, parsed?.message, parsed?.delta]) {
+          mergeUsage(this.usage, usageFromJson("anthropic", source));
+        }
       } else {
         mergeUsage(this.usage, usageFromJson("openai", parsed));
       }
--- a/services/gateway/test/gateway.test.ts
+++ b/services/gateway/test/gateway.test.ts
@@ -374,6 +374,15 @@ describe("gateway", async () => {
       await db.select().from(llmRequests).where(eq(llmRequests.virtualKeyId, setup.keyId))
     )[0];
     expect(row?.outputTokens).toBe(1_000_000);
+    // Anthropic reports streamed input usage in message_start, output in
+    // message_delta. Both have to reach the row and the budget counter:
+    // claude-sonnet-5 is $3/1M input + $15/1M output, so 1M each is 1800 cents.
+    expect(row?.inputTokens).toBe(1_000_000);
+    expect(row?.costCents).toBe(1800);
+    const counter = (
+      await db.select().from(spendCounters).where(eq(spendCounters.budgetId, setup.budgetId))
+    )[0];
+    expect(counter?.spentCents).toBe(1800);
   });
 
   it("2b. shutdown drains post-response metering before closing Postgres", async () => {
--- a/services/gateway/test/usage.test.ts
+++ b/services/gateway/test/usage.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+import { UsageTee } from "../src/usage.js";
+
+const MESSAGE_START =
+  'event: message_start\ndata: {"type":"message_start","message":{"usage":{"input_tokens":1000,"cache_read_input_tokens":40,"cache_creation_input_tokens":25}}}\n\n';
+const CONTENT_DELTA =
+  'event: content_block_delta\ndata: {"type":"content_block_delta","delta":{"text":"ok"}}\n\n';
+const MESSAGE_DELTA =
+  'event: message_delta\ndata: {"type":"message_delta","delta":{"usage":{"output_tokens":700}}}\n\n';
+
+describe("UsageTee anthropic streaming usage", () => {
+  it("meters input and cache tokens reported by message_start", async () => {
+    const tee = new UsageTee("anthropic");
+    await feed(tee, [MESSAGE_START, CONTENT_DELTA, MESSAGE_DELTA]);
+
+    expect(tee.usage).toEqual({
+      inputTokens: 1000,
+      outputTokens: 700,
+      cacheReadTokens: 40,
+      cacheWriteTokens: 25,
+    });
+  });
+
+  it("keeps message_start input tokens when message_delta reports only output", async () => {
+    const tee = new UsageTee("anthropic");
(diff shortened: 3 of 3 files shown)


jungle-lethanh:
I followed the EOF case separately from this fix. #148 looks good for the message_start and merge changes, but there’s one edge case left in the shared SSE path.
If the last usage frame contains valid JSON but the stream closes before the final blank line, the response still passes through, but that usage never gets counted.
This looks separate from #54. Since SSE normally treats that last frame as incomplete, I’m wondering which behavior Facility actually wants here: follow the protocol strictly, or still recover the usage because it affects budget tracking.
I can add a focused test once that behavior is clear.

haiderysfzi:
Good catch, and I reproduced it. With the last frame valid but unterminated, the bytes still pass through byte-exact and usage comes back with `outputTokens: 0`. The cause is in `parsePending`: the trailing remainder is handed to `parseJsonText`, so it gets `JSON.parse`d with the `event:`/`data:` prefix still attached and throws. `parseSseFrame` is never applied to the final partial frame. It predates this PR and affects the OpenAI path the same way, since that part of the tee is provider-agnostic.

On which behaviour Facility wants, I would argue for recovering it.

The repo already takes that position elsewhere: in `writeMetering`, when a request errored and `providerMayHaveCharged` with a measured cost of zero, the cost falls back to the estimate. That is the same judgement, which is that when the provider may have billed you count something rather than nothing.

Strict SSE framing governs what you forward, and the tee already forwards untouched, so counting a frame it fully received breaks no contract with the caller. `UsageTee` is a meter rather than an EventSource client.

And the consequence is not a cosmetic metric: `spend_counters` is what `hardBudgetBlock` enforces against, so a dropped usage frame is free tokens against a hard cap, and a truncated stream is exactly when you least want that. It stays conservative either way, since only valid JSON recovers and a genuinely truncated body still parses to nothing and counts nothing.

The change is small: route the flush remainder through `parseSseFrame` when it contains `data:` lines, falling back to `parseJsonText` otherwise. Happy to add it here if you would rather have one round, but I would lean toward keeping it out of #54 so this diff stays about the `message_start` source. Tell me which you prefer and I will push it, or leave it to you if you already have the test written.

adrian-lorenzo (approved):

haiderysfzi:
Thanks! @adrian-lorenzo. Happy to dig into the codebase. CI workflows still need a maintainer to approve the run since it's my first contribution, could someone hit the approve button when you get a chance?

ophiocus:
Not asking to reopen anything — the production fix is right and the approval looks right to me. This is about the test corpus, and it's a follow-up rather than a blocker.

`api.anthropic.com` puts `usage` at the **top level** of `message_delta`, as a sibling of `delta`; `delta` carries only the stop fields. That's typed in this repo's own dependency — `@anthropic-ai/sdk@0.112.5`, `resources/messages/messages.d.ts`:

```ts
export interface RawMessageDeltaEvent {
    delta: RawMessageDeltaEvent.Delta;   // stop_reason, stop_sequence, …
    type: 'message_delta';
    usage: MessageDeltaUsage;            // <- sibling, not inside delta
}
```

Two places still assume the nested shape. One is `anthropicSseBody()`, which predates this PR — it's been that way since the original gateway commit, so I've sent a separate PR for it against `main` (#181) rather than adding to this branch. The other is the comment added here in `usage.ts`:

> `message_delta` carries output tokens under `delta.usage`

That one's worth a two-word edit before it lands, since it's the explanation the next person will read.

Why it matters beyond tidiness — two mutations to `usage.ts` on this branch, running `test/usage.test.ts`:

| mutation | with these fixtures |
|---|---|
| drop `parsed?.delta` (dead against the real wire) | 4 of 5 **fail** |
| drop `parsed` (the source that meters every streamed output token) | 4 of 5 still **pass** |

So the streaming tests are load-bearing on a frame that never arrives, and the source that actually meters streamed output is pinned only by the non-streamed case. `mergeUsage`'s last-wins handles the real shape correctly already — there's just nothing holding it in place.

One practical note: I checked that my stub PR and this branch don't collide. Merged together locally, the gateway suite is 57/57, and your new `inputTokens` / `costCents` / `spentCents` assertions all still pass. If it's easier, ignore mine and fold the stub change in here instead — happy either way.

I also have the fixture corrections for `usage.test.ts` plus a cumulative-restatement case if you want them, but that's your call, not a request.

Issue #104, opened 2026-08-07
Find a story by number or title from the Stories board


javiertoledo:
## The problem

The Stories board has exactly one filter: the seven stage chips (`apps/web/app/(app)/projects/[projectId]/stories/page.tsx:78-110`). To reach a specific story you scroll a board that renders every open story in the project, grouped into seven collapsible sections. There is no way to type a number, or a few words of a title, and land on it.

The ⌘K palette does not help. It navigates to *sections* — projects, stories, sessions, approvals, settings (`apps/web/components/shell/cmdk.tsx:29-79`) — and its own docstring says verbs and objects *"join as surfaces land"*. Stories are a surface that has landed.

## What I propose

**A filter box on the board**, filtering the already-fetched pipeline client-side on number, title, repository and assignee. `api.pipeline(projectId)` (`stories/page.tsx:41`) already returns every visible story with all of those fields, so this is a client component over data in hand: no request, instant feedback per keystroke, and it composes with the stage chips rather than replacing them.

**And a number that jumps.** When the query is a bare number (`912` or `#912`), offer *go to story #912* as the first result and navigate to `/projects/{id}/stories/912?repoId=…` on Enter.

That second half matters because of a real boundary. `/pipeline` returns open stories plus those shipped in the last seven days (`assembleProjectStories`, `services/api/src/routes/v1/github.ts:987`), so a filter over the board **cannot** reach an older closed story — while the story endpoint resolves any number in the mirror (`assembleSelectedStories`, `:1051`). Typing a number should therefore work regardless of the window; typing words honestly cannot, and the empty state should say *"not on the board — older stories are searchable by number"* rather than imply the story does not exist.

With several connected repositories the jump needs a `repoId`. When the number is not on the board, either resolve without one and let the endpoint's 409 (`ambiguous_story_number`) drive a repository picker, or offer one entry per repo up front.

**Then feed the palette.** Once the board has a matcher, ⌘K gaining `story #N / title` entries for the current project is the same predicate reused — and that is the answer to *"from anywhere, take me to that story"*.

## What I considered instead

- **A server-side search endpoint.** The right long-term answer for full history — `gh_issues` is already indexed on `(org, project, gh_updated_at)` and a trigram index on `title` would be cheap — but it is not what makes the common case fast. The story you are looking for is almost always on the board, and the board is already in memory.
- **Filtering through the URL, like the stage chips.** A round trip per keystroke for data the client already holds. A stage chip is a link because it is a shareable view; a search box is not.

[...]

PR #150, opened 2026-08-14
feat(web): add story filtering and number jump

Bobbyap72:
web: add title filter and number-jump on the Stories board (#104)

Adds a client-side filter box on the Stories page that narrows
visible stories by title match, and lets a bare story number jump
straight to that story. Numbers already on the board resolve
locally; numbers outside the board's 7-day window fall back to
GET /v1/projects/:projectId/stories/:number.

Known gap: the 409 ambiguous_story_number response doesn't
currently return the candidate repos, so ambiguous numbers show
a message rather than a repo picker. Extending that response is
a natural follow-up, kept out of this PR to stay scoped to
apps/web.

### DIFF
--- a/apps/web/app/(app)/projects/[projectId]/stories/page.tsx
+++ b/apps/web/app/(app)/projects/[projectId]/stories/page.tsx
@@ -1,11 +1,10 @@
 import { cx, Eyebrow, StatusDot } from "@facility/ui";
 import Link from "next/link";
 import type { ReactNode } from "react";
-import { IssueRow } from "@/components/issues/issue-row";
 import { SyncIssuesButton } from "@/components/issues/sync-button";
 import { ErrorNotice, Offline } from "@/components/offline";
-import { StageSection } from "@/components/project/stage-section";
 import { LiveRefresh } from "@/components/shell/live-refresh";
+import { StoriesBoard } from "@/components/story/stories-board";
 import { api } from "@/lib/api";
 import type { PipelineStageKey, PipelineStageKind, PipelineStageState } from "@/lib/pipeline";
 import { pipelineStageStateLabel, pipelineStories } from "@/lib/pipeline";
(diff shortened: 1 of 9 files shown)


adrian-lorenzo (changes requested):
Thanks for the contribution!

Story-number jumps need to handle duplicates before this can merge. `localMatch` uses the first matching story, so if two repositories have the same number, pressing Enter can silently open the wrong one. Off-board duplicates also leave the user without a working way to choose a repository. Please make ambiguous matches explicit and add tests covering both local and API ambiguity.

The branch also conflicts with `main`. When rebasing, please preserve the Builder plan policy by passing `builderPlanRequired` through `StoriesBoard` to each `IssueRow`. Please also distinguish API failures from a genuine 404, fix the import ordering, and update the title and commit subject to an allowed Conventional Commit such as `feat(web): add story filtering and number jump`.

With those changes, we can approve and merge it!

Bobbyap72:
Thanks for the thorough review — pushed a rebased, squashed commit addressing all six points:

1. **Ambiguity handling**: both local and off-board duplicates are now explicit, not silently resolved to the first match. Locally, `resolveLocalJump` reports `ambiguous` if two repos on the board share a number. Off-board, the `409` response from `/stories/:number` now includes the candidate repos (`repoId`/`repoOwner`/`repoName`/`storyType`) — the UI renders a real picker instead of a dead-end message. Covered by `apps/web/test/story-jump.test.ts`.
2. **Rebased** onto current `main`.
3. **`builderPlanRequired`** now threaded through `StoriesBoard` → `IssueRow`.
4. **API failures vs. genuine 404**: `interpretJumpResponse` now distinguishes a real 404 from a 5xx/network failure — the UI says "couldn't check" rather than implying non-existence.
5. **Import ordering** fixed via `pnpm exec biome check --write`.
6. **Title/commit** updated to `feat(web): add story filtering and number jump`.

One note on scope: this does touch `services/api` now (the 409 response body), beyond the original `apps/web`-only plan — necessary since building a real picker needed the candidate data the API wasn't returning.

Also flagging: `pnpm --filter @facility/api typecheck` currently reports 43 pre-existing errors on `main` (missing `builderPlanPolicy`/`workspaceBaseSha` types, traced to #205). Confirmed unrelated by running the identical typecheck against a clean `upstream/main` checkout — same 43 errors, same files/lines, before any of this branch's changes were applied. Not touched here to keep this PR to one intent.

Bobbyap72:
Confirmed the title and commit subject landed correctly:

    $ gh pr view 150 --repo theam/facility --json title,commits --jq '.title, .commits[-1].messageHeadline'
    feat(web): add story filtering and number jump
    feat(web): add story filtering and number jump

Ready for another look whenever you get a chance.

Bobbyap72:
Good Morning Adrián,
Thanks for that.

Pushed a rebased, squashed commit addressing all six points, plus a couple
of follow-up comments on the PR with details and verification output.

Ready for your review, whenever it works for you.

Thanks,
Bobby AP

adrian-lorenzo (changes requested):
Thanks for the thoughtful revision!

The original issues are addressed, but the off-board lookup still has a race. If someone submits one number and then changes or submits another before the first request completes, the older response can arrive last and navigate them to the wrong story or overwrite the current status. Please cancel superseded requests or ignore their responses, and add a component-level regression covering out-of-order responses.

Please also announce the asynchronous lookup result through an `aria-live` region, use links for the repository choices, and extend the API integration test to verify the 409 response’s `matches` data that the picker now depends on.

With those small improvements, this should be ready to approve.

Bobbyap72:
Dear Adrián,
Good Evening!

Thanks for the follow-up review. Pushed fixes for all four points, plus a
couple of comments on the PR with the details and verification.

Please check it out, at your convenience.

Thanks,
Bobby AP

Bobbyap72:
Hello Adrian,

Hope you are fine.

Did you get a chance to look into the results? Please let me know if
anything else is needed.

Looking forward to it.

Kind regards,
Bobby

Bobbyap72:
Hi Adrián,

Just following up on PR #150 whenever you get a chance to take another
look,  no rush at all, just wanted to check in.

I will be travelling to the Netherlands on 12th Sep, that is my base by the
way.

Hope things are going well on your end.

Thanks,
Bobby AP

PR #159, opened 2026-08-14
fix(api): treat empty optional env vars as unset

### LINKED ISSUE #158: Fresh `pnpm dev` crashes the API and worker: blank VERCEL_* lines in the generated .env fail config validation
## What happens

On a clean clone, the quickstart boot crashes the API and the worker while the web app, gateway, and docs come up fine — so the stack looks healthy unless you probe the ports.

```bash
git clone https://github.com/theam/facility.git && cd facility
corepack enable
pnpm dev
```

```
@facility/api:dev:worker: ZodError: [
  { "code": "too_small", "path": ["VERCEL_TOKEN"], "message": "Too small: expected string to have >=1 characters" },
  ... same for VERCEL_OIDC_TOKEN, VERCEL_TEAM_ID, VERCEL_PROJECT_ID
]
    at readConfig (services/api/src/config.ts:187)
```

Both entry points die: `services/api/src/dev.ts` and `services/api/src/worker.ts`. Environment: WSL2 Ubuntu, Node 22.21.1, pnpm 11.20.0, Docker 29.2.1, `main` @ `ae68401`.

## Root cause

`.env.example` ships the optional Vercel sandbox variables as blank assignments:

```dotenv
VERCEL_TOKEN=
VERCEL_OIDC_TOKEN=
VERCEL_TEAM_ID=
VERCEL_PROJECT_ID=
```

`pnpm dev` copies them verbatim into `.env`, and dotenv delivers `KEY=` as an **empty string**, not `undefined`. The schema declares them as `z.string().trim().min(1).optional()` (`services/api/src/config.ts:74-77`) — `.optional()` only tolerates `undefined`, so empty-present fails `min(1)` and `readConfig` throws.

The template and the schema disagree about what "unset" looks like. Notably, the same file already resolves this for other optional vars: `OptionalUrl`, `OptionalGithubOrganization`, and `FACILITY_PREVIEW_SURFACE_TOKEN` all preprocess blank strings to `undefined` (`config.ts:11-37`). The `min(1).optional()` group (`VERCEL_*`, `FACILITY_AWS_CODEBUILD_*`, `PACKAGE_REGISTRY_TOKEN`) is the outlier — it just happens that only the `VERCEL_*` keys appear blank in `.env.example`, which is why only they crash a fresh boot.

## Two possible fixes

**Option A — schema-side (my recommendation):** extend the file's existing blank-to-undefined idiom to the `min(1).optional()` group:

```ts
const OptionalNonEmpty = z.preprocess(
  (value) => (typeof value === "string" && value.trim() === "" ? undefined : value),
  z.string().trim().min(1).optional(),
);
```

This fixes the class, not the instance: `KEY=` means "unset" for every optional non-empty var, existing `.env` files heal without edits, and the behavior matches how `facility doctor` already reasons about these values (truthiness — `services/api/src/doctor.ts:121-133` treats empty and absent identically).

**Option B — template-side:** comment out the four `VERCEL_*` lines in `.env.example` and keep the strict schema. Preserves strict-presence semantics, but anyone re-adding `KEY=` re-breaks the boot, and already-generated `.env` files stay broken.

For what it's worth, I don't think Option A changes any intentional guarantee: there is no cross-field validation requiring the tokens when `FACILITY_SANDBOX_DRIVER=vercel` (absent tokens already boot fine today), so the current `min(1)` only rejects one spelling of "not configured" while accepting the other. No existing test pins the empty-string rejection.

I have a small PR ready implementing Option A with unit tests (empty, whitespace-only, set, and absent cases) — happy to rework it toward Option B if you prefer the strict semantics.

stringbitking:
Fixes #158 — a fresh clone's `pnpm dev` crashes the API and worker because `.env.example` ships the `VERCEL_*` keys as blank assignments (`KEY=`), dotenv delivers those as empty strings, and the `min(1).optional()` declarations in `services/api/src/config.ts` reject empty-present values.

## What this does

Implements **Option A** from the issue: extends the blank-to-undefined preprocess idiom the file already uses for `OptionalUrl`, `OptionalGithubOrganization`, and `FACILITY_PREVIEW_SURFACE_TOKEN` to the whole optional non-empty group — `VERCEL_TOKEN`, `VERCEL_OIDC_TOKEN`, `VERCEL_TEAM_ID`, `VERCEL_PROJECT_ID`, `FACILITY_AWS_CODEBUILD_PROJECT`, `FACILITY_AWS_CODEBUILD_CACHE_BASE_LOCATION`, and `PACKAGE_REGISTRY_TOKEN`.

A blank or whitespace-only value now parses as unset (matching how `facility doctor` already reasons about these values via truthiness), while a value that is actually set must still be non-empty after trimming. Existing generated `.env` files heal without edits.

## Why this doesn't change an intentional guarantee

- There is no cross-field validation requiring the tokens when `FACILITY_SANDBOX_DRIVER=vercel` — absent tokens already boot fine today, so the current `min(1)` only rejected one spelling of "not configured" while accepting the other.
- No existing test pinned the empty-string rejection; the Vercel config test covers trimming and OIDC-over-token precedence, both preserved.
- `facility doctor` remains the fail-fast surface for "driver selected but not configured".

Happy to rework toward Option B (comment the blank lines out of `.env.example`, keep the strict schema) if you prefer strict-presence semantics.

## Tests

Two new cases in `services/api/test/config.test.ts`:
- the exact `.env`-template shape (all seven keys blank) boots and resolves every value to `undefined`;
- a whitespace-only value resolves identically to the key being absent.

`vitest run test/config.test.ts`: 16 passed. `tsc --noEmit` and `biome check` clean on the touched files.


adrian-lorenzo (approved):
Thanks for the contribution!

Issue #70, opened 2026-08-04
Show the signed-in GitHub user, and assign them on the issue and the PR


javiertoledo:
## The problem

Facility signs you in with GitHub, but never shows you as your GitHub self and never tells GitHub you are working on something.

**Your GitHub login is stored and never read.** `auth.ts:241-248` writes `user_identities.login` when the identity is first linked, and `:265-273` refreshes it on every subsequent sign-in — precisely so it stays current. Nothing anywhere queries it back. The reason is one layer up: `PrincipalSchema` (`services/api/src/routes/v1/shared.ts:117-126`) has `type, id, orgId, userId, email, name, projectId, permissions` and no login, so the field cannot reach the client even if it were resolved. `apps/web/components/shell/topbar.tsx:24` consequently renders:

```tsx
<span className="font-mono text-[11px] text-(--dim)">{me.principal.email}</span>
```

The signed-in user sees an email address where they expect `@their-handle`. `users.avatar_url` is populated on every login too (`auth.ts:257-264`, sourced from GitHub at `auth/identity-provider.ts:144`) and is likewise never rendered.

**Triggering an agent leaves no trace of who did it on GitHub.** `POST /v1/projects/:projectId/issues/:number/trigger` (`services/api/src/routes/v1/github.ts:526-611`) creates the run and posts an acknowledgement comment via `ackIssueQueued`, but assigns nobody. Nor does the PR that run later opens (`services/api/src/sandbox/orchestrator.ts:791`). So on GitHub, work someone deliberately started looks unowned.

## What I propose

**Surface the identity.** Add `githubLogin` and `avatarUrl` to `PrincipalSchema` and to the `Principal` type (`services/api/src/types.ts:4-13`), resolved via a left join on `user_identities` (`provider = 'github'`) in both user paths of `resolvePrincipal` — the OAuth token path and the session cookie path in `services/api/src/app.ts`. The API-key path should not gain them; a key is not a person. `user_identities_user_provider_uidx` is `unique(user_id, provider)`, so the join cannot multiply rows.

Adding the field to the schema is load-bearing, not cosmetic: `fastify-type-provider-zod` serialises responses *through* the schema, so a field set by `resolvePrincipal` but missing from `PrincipalSchema` is silently stripped. The change requires regenerating `packages/sdk/openapi.json` and `schema.d.ts` in the same commit — `api.test.ts` asserts the committed OpenAPI is byte-identical to the live schema.

[...]

PR #161, opened 2026-08-15
feat(web): render GitHub assignees on stories

### LINKED ISSUE #101: Show who owns a story — render the assignees Facility already mirrors
## The problem

Every story on the board is unowned as far as the interface is concerned. Nowhere in `apps/web` is `assignees` read — not on the row, not on the story header, not on the pipeline board.

The data has been there all along, and #70 just made it meaningful:

- `gh_issues.assignees` is mirrored on every webhook and every sync (`packages/db/src/schema.ts:234`, `services/api/src/github/issues-sync.ts:281`);
- it is assembled into the story (`services/api/src/pipeline.ts:178`);
- it is serialised on **both** responses the web already fetches — `PipelineStorySchema.assignees` (`services/api/src/routes/v1/github.ts:189`) and `StoryDetailSchema.assignees` (`:228`);
- and since #70, triggering an agent assigns the triggering user on the issue and on the PR the run later opens (`services/api/src/routes/v1/github.ts:944`, `services/api/src/sandbox/orchestrator.ts:891`).

So Facility writes ownership to GitHub, mirrors it back, ships it over the wire — and drops it at the last inch. On a board where several people dispatch agents against the same repository, *"who is on this?"* is unanswerable without opening GitHub.

## What I propose

Render the assignee on the story row and in the story header, in the existing grammar: `@login`, mono, `--dim`, beside the labels and the timestamp in `IssueRow` (`apps/web/components/issues/issue-row.tsx:84-92`). Several assignees: the first plus `+N`. Unassigned: render nothing, not a placeholder — the rows are dense and unassigned is the common case in Backlog.

Two things that make it more than decoration, worth the same slice:

- **Avatars.** `users.avatar_url` exists for Facility users, but a GitHub assignee need not be one; `https://github.com/{login}.png?size=40` needs no new column and no new sync. Where a deployment must not let the browser reach github.com, the initial-letter fallback the topbar already uses is the answer.
- **A "mine" filter on the Stories board**, next to the stage chips, matching against `me.principal.githubLogin` (added by #70, `services/api/src/routes/v1/shared.ts:123`). That is what turns a rendered field into a working view.

## What I considered instead

- **Showing the last run's `triggeredBy` as the owner.** That is who dispatched the most recent agent — often, but not always, who owns the story — and it is empty for anything nobody has dispatched yet. The GitHub assignee is the answer the rest of the world already agrees on.
- **A Facility-native owner field.** A second source of truth for something GitHub models, with no writer, no sync path and no reconciliation story.

The change is `apps/web`-only: the data is already on the wire, and `apps/web/test/pipeline-story.test.ts` is where the row's rendering rules are pinned.

Related: #70, #18. Findings are against `5712bf8`.

Nxssie:
## Summary

Closes #101.

Facility already mirrors `gh_issues.assignees` end to end — webhook sync,
`services/api/src/pipeline.ts`, and both `PipelineStorySchema.assignees` and
`StoryDetailSchema.assignees` — but `apps/web` never read the field, so
ownership was invisible without opening GitHub.

- Render `@login` assignee chips on the story row (`IssueRow`) and the story
  detail header, in the existing label/timestamp grammar. Several assignees
  collapse to the first plus `+N`; unassigned renders nothing (Backlog's
  common case), not a placeholder.
- Add `AssigneeChip`: GitHub avatar (`github.com/{login}.png`, no new sync
  column needed) with an initial-letter fallback for deployments where the
  browser can't reach `github.com`.
- Add a "mine" filter on the Stories board, next to the stage chips, matching
  against the signed-in user's `githubLogin` (from #70).

## Test plan

- [x] `bunx tsc --noEmit` — clean
- [x] `bunx vitest run` (apps/web, full suite) — 67/67 passing, including new
      coverage in `pipeline-story.test.ts` for the unassigned, +N, and
      avatar/fallback cases
- [x] `bunx biome check` on touched files — clean
- [x] Manual check in a browser (not run in this environment)

### DIFF
--- a/apps/web/app/(app)/projects/[projectId]/stories/[number]/page.tsx
+++ b/apps/web/app/(app)/projects/[projectId]/stories/[number]/page.tsx
@@ -2,6 +2,7 @@ import { Eyebrow, PillTag, StatusDot } from "@facility/ui";
 import Link from "next/link";
 import { notFound } from "next/navigation";
 import { CiStatusLink } from "@/components/ci-status";
+import { AssigneeChip } from "@/components/issues/assignee-chip";
 import { Markdown } from "@/components/markdown";
 import { ErrorNotice, Offline } from "@/components/offline";
 import { LiveRefresh } from "@/components/shell/live-refresh";
@@ -164,6 +165,9 @@ export default async function StoryPage({
             >
               {label}
             </span>
+          ))}
+          {story.assignees.map((login) => (
+            <AssigneeChip key={login} login={login} />
           ))}
           <a
             href={story.htmlUrl}
--- a/apps/web/app/(app)/projects/[projectId]/stories/page.tsx
+++ b/apps/web/app/(app)/projects/[projectId]/stories/page.tsx
@@ -35,9 +35,9 @@ export default async function ProjectStoriesPage({
   searchParams,
 }: {
   params: Promise<{ projectId: string }>;
-  searchParams: Promise<{ stage?: string; status?: string }>;
+  searchParams: Promise<{ stage?: string; status?: string; mine?: string }>;
 }) {
-  const [{ projectId }, { stage, status }] = await Promise.all([params, searchParams]);
+  const [{ projectId }, { stage, status, mine }] = await Promise.all([params, searchParams]);
   const [pipelineResult, me] = await Promise.all([api.pipeline(projectId), api.me()]);
 
   if (!pipelineResult.ok && pipelineResult.offline) return <Offline />;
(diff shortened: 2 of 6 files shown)


Nxssie:
Pushed a follow-up commit (a98dd21) after manually walking through this on a local instance:

- The **mine** filter was originally placed inline between "all" and the stage tabs. Visually that read as if it were part of the same single-select group ("all", "Shipped", "In review", ...), when it's actually an independent axis (assignee, not stage) — the two can be active at once (e.g. "all stages, only mine").
- Moved it to the end of the filter row, set off with a vertical divider, so it reads as its own filter rather than a peer of the stage tabs.

No behavior change to the filtering logic itself, just the grouping/placement.

adrian-lorenzo:
Thanks for the contribution!

There are two ideas here worth preserving: visually separating the “mine” filter from the stage filters, since they are independent axes, and showing the complete assignee list on the story detail while keeping the board row compact.

#171 now covers the underlying assignee rendering and filter behavior, while #175 covers avatars. Please reframe this PR as a focused follow-up containing only those distinct UI improvements, with the current Builder gate wiring from `main` preserved.

That would give this contribution a clear, useful scope without duplicating the other work.

Nxssie:
Thanks for the review, and agreed on the scope.

I'll wait for #171 to land, then rebase this branch onto `main` and narrow it to the two UI changes that aren't covered there:

1. **"mine" as its own filter axis** — moved to the end of the filter row behind a divider, so it doesn't read as a peer of the stage tabs. Filtering logic stays as #171 ships it (`ownedBy`, `mineFilterOn`, `boardHref`).
2. **Full assignee list on the story detail header** — every `@login` rather than the first plus `+N`, while the board row keeps the compact form from `storyOwner`.

Everything else goes: the `AssigneeChip` component, the avatar fallback, and my own presenter/tests, since #171 and #175 cover them. The Builder gate wiring on `main` stays untouched.

I'll retitle the PR to match the reduced scope once it's rebased. If you'd rather I close this and open a fresh follow-up against the merged #171, that works for me too.

PR #164, opened 2026-08-16
fix(sandbox): resolve codex-prefixed pull request roles

### LINKED ISSUE #163: fix(sandbox): codex-prefixed review and repair agents clone the default branch
## What I found

`githubPullRequestMode` is the only role predicate in `services/api/src/sandbox/orchestrator.ts` that does not strip the `codex-` prefix from `runs.mode`:

```ts
// orchestrator.ts:2350
function repairPullRequestMode(mode: string) {
  return ["address_review", "ci_doctor"].includes(mode.replace(/^codex-/, "").replace(/-/g, "_"));
}

// orchestrator.ts:2364
function readOnlyRepositoryMode(mode: string) {
  return ["architect", "review", "security_sweep"].includes(
    mode.replace(/^codex-/, "").replace(/-/g, "_"),
  );
}

// orchestrator.ts:2370  — no prefix strip
function githubPullRequestMode(mode: string) {
  return ["review", "address_review", "ci_doctor"].includes(mode.replace(/-/g, "_"));
}
```

It is load-bearing at `orchestrator.ts:1733`, which decides whether the run bundle receives the pull request's branch:

```ts
const checkoutBranch = githubPullRequestMode(run.mode) && githubBranch ? githubBranch : null;
const expectedHeadSha = repairExpectedHeadSha(run.mode, run.trigger);
```

and then `branch: checkoutBranch ?? repo.defaultBranch`.

## Why it triggers

`runs.mode` is the agent's name (`routes/v1/runs.ts:246`), and Codex agents carry a `codex-` prefix by convention — `packages/db/src/seed.ts:52-53` seeds `codex-architect` and `codex-builder`. A project that follows that convention and adds a Codex reviewer or repair agent hits this.

## Impact

- **`codex-review`** — `checkoutBranch` is null and no head is pinned, so the agent clones the **default branch** and reviews the wrong tree while reporting on the pull request. This one is silent.
- **`codex-address-review` / `codex-ci-doctor`** — `repairExpectedHeadSha` *does* strip the prefix, so the bundle asks the runner to clone the default branch and then verify the pull request's head. The runner's check at `runner/src/index.ts:460-471` throws `repository_head_sha_mismatch` and the run fails before provisioning.

So the two predicates disagree about the same run: one says "this is not a pull-request role", the other says "pin this PR's head".

Latent on a default install because the seed ships only `codex-architect` and `codex-builder`.

## Suggested fix

Strip the prefix exactly as the sibling predicates do. Happy to open the PR — I have the one-line change plus a regression test that fails before it and passes after.

Found while reading the orchestrator; I have not run this against a live Codex agent, so please sanity-check the impact description against how you expect projects to name Codex agents.

igoraboim:
Closes #163.

## What changes

`githubPullRequestMode` now strips the `codex-` prefix from `runs.mode`, exactly as `readOnlyRepositoryMode` and `repairPullRequestMode` already do. After this, a Codex-engine reviewer or repair agent resolves to the same governed role as its Claude counterpart and receives the pull request's branch in its run bundle.

## Why

`runs.mode` is the agent's name (`routes/v1/runs.ts:246`) and Codex agents carry a `codex-` prefix by convention — the seed ships `codex-architect` and `codex-builder` (`packages/db/src/seed.ts:52-53`). `githubPullRequestMode` was the only role predicate that did not normalise that prefix, and it decides whether the bundle gets the PR branch (`orchestrator.ts:1733`).

A project that follows the naming convention and adds a Codex reviewer or repair agent gets `checkoutBranch = null` and clones the default branch:

- `codex-review` reviews the default branch while reporting on the pull request — silent.
- `codex-address-review` / `codex-ci-doctor` also pin an admitted head through `repairExpectedHeadSha`, which *does* strip the prefix. The bundle then asks the runner to clone the default branch and verify the PR's head, so the run fails at `repository_head_sha_mismatch` (`runner/src/index.ts:460-471`) before provisioning.

The two predicates disagreed about the same run. The second test added here pins that agreement rather than just the prefix handling, so the invariant survives a future role being added to one list and not the other.

The predicate is exported to make it testable, following the precedent of `repairExpectedHeadSha` (exported and covered in `test/sandbox.test.ts`).

## Verification

- `services/api` full suite: **234 passed, 16 skipped** (`npx vitest run --fileParallelism=false`). The skips are the Docker/Postgres-backed tiers — see the caveat below.
- Targeted: `test/sandbox.test.ts`, `test/orchestrator-checks.test.ts`, `test/ci-doctor-policy.test.ts`, `test/address-review-policy.test.ts`, `test/sandbox-capabilities.test.ts` — 69 passed in the four policy suites.
- `node guards/run.mjs` — 2 guards ran, 0 failed.
- `npx tsc --noEmit` in `services/api` — clean. `npx biome check` on both changed files — clean.
- **The new tests were confirmed to fail without the fix.** I reverted `githubPullRequestMode` to its previous body and re-ran: both new cases fail (`expected false to be true`, at the `codex-` assertions and at the branch/head agreement check), then pass again with the fix restored. They are a regression pin, not a restatement.

Caveat, stated plainly: I could not run `pnpm verify` end to end — Docker is unavailable on this machine, so the isolated-database integration tier and the sandbox E2E could not execute locally. Everything above is what I could actually run. I also have not exercised this against a live Codex agent; the impact description is traced from the code path, and I would value a maintainer sanity-check on how you expect projects to name Codex agents.

- [x] `pnpm verify` passes locally — **not run**, no Docker available; see the caveat above
- [x] Behaviour verified beyond the test suite — verified by reverting the fix and confirming the new tests fail, and by tracing the bundle through `buildRunBundle` to the runner's `repository_head_sha_mismatch` check
- [x] Documentation updated, or no user-facing change — no user-facing change


UsamaIslam:
Hit this exact bug independently while auditing the runner/gate boundary, before finding this PR — same fix (`mode.replace(/^codex-/, "").replace(/-/g, "_")`, matching `repairPullRequestMode`). I verified it two ways: a full truth table across all mode variants (bare and `codex-`-prefixed review/address-review/ci-doctor/builder/architect/security-sweep — only the three PR-bound codex- modes change), and reverting just the regex while keeping the export, which reproduces exactly one failure: `codex-review` expected `true`, got `false`. Consistent with what you found — independent +1.

adrian-lorenzo (approved):

Issue #147, opened 2026-08-13
The base commit a run started from is never recorded, so a stale plan is invisible at Gate 1


goyaniparas4217:
Facility records which repository, issue and pull request a run touched, and where the work landed. It does not record the commit the work started from.

## What is true today

**Every SHA the platform stores is a head, written after the work exists.** `run_deliveries.expected_head_sha` is the delivery branch head after the push (`services/api/src/sandbox/orchestrator.ts:969`). `gh_pull_requests` stores `head_sha` next to `base_ref` (`packages/db/src/schema.ts:274-276`). `preview_sandboxes.commit_sha` is the commit that got built (`:1095-1097`). `runs` has no commit column (`:469-511`), and neither does `proposals` (`:826-847`).

One function shows the gap. `runDeliveryRefMismatch` (`orchestrator.ts:1281-1290`) compares `actual.baseRef` against `expected.baseBranch`, which are branch names, then compares `actual.headSha` against `expected.expectedHeadSha`, which are commits. **The head is checked by commit. The base is checked by name.** A base branch that keeps its name and moves forty commits passes this check.

One line reads like it already settles this, so it is worth naming. `apps/docs/docs/reference/architecture.md:109` says "owner, repository, base, or SHA drift blocks delivery". Against the check above, "base" means the ref name and "SHA" means the head. The sentence is correct. It does not cover a base that moved.

**Here is the part that makes this worth filing: the runner already works out the base commit, then throws it away.** `shipGitChanges` runs `rev-parse origin/<base>` (`runner/src/index.ts:1762`) and uses it to create the branch ref and as the concurrency guard. It returns `{ branch, headSha, changed }` (`:1807-1817`). The base is computed on every delivery and never leaves the function.

A plan lives in the issue thread, and acceptance is invoking `/builder`. Any amount of time can pass between the two, and nothing records whether `main` moved. Four things follow:

- A stale plan looks the same as a fresh one at Gate 1.
- Two stories in flight have no recorded order.
- `outcomes` counts `review_rounds` and `fixup_commits` (`schema.ts:1050-1080`) but cannot tell rework caused by another story landing first from any other rework.
- A receipt's `github` block is `owner`, `repo`, `issue`, `pr`, `actor_sha256` (`packages/core/src/receipts.ts:48-56`). A reader can see which issue an agent worked from, but not the tree it was reading.

`AGENTS.md:51` already advises requiring an up-to-date branch or a merge queue. That is the mitigation, and it lives outside the product.

[...]

PR #166, opened 2026-08-16
feat(runs): record base SHA provenance

jungle-lethanh:
This PR does not compare `workspaceBaseSha` with the live branch during plan acceptance and does not alter Gate 1 or `/builder` dispatch; it provides the durable input required for that follow-up behavior.
## What changes

Records the Git base commit for repository-backed runs and deliveries.

- Adds nullable `runs.workspace_base_sha`.
- Adds nullable `run_deliveries.base_sha`.
- Captures the workspace `HEAD` after checkpoint restoration and before agent execution.
- Adds an authenticated runner endpoint for recording workspace provenance.
- Makes workspace provenance immutable:
  - First write succeeds.
  - Identical retries succeed.
  - Conflicting retries return `409 workspace_base_mismatch`.
- Preserves the delivery base already calculated by `shipGitChanges`.
- Adds optional `github.base_sha` receipt provenance.
- Exposes `workspaceBaseSha` through the public Run API and generated SDK.

Workspace and delivery bases remain separate:

| Field | Meaning |
| --- | --- |
| `workspaceBaseSha` | Commit presented to the agent |
| Delivery `baseSha` | Base of the published change set |
| `expectedHeadSha` | Published delivery head |

For delivery receipts, `github.base_sha` uses the delivery base. Runs without a delivery fall back to the prepared workspace base.

## Why

Facility records the resulting head commit, but not the commit from which the agent started.

A branch name such as `main` can remain unchanged while its commit advances. Without the original base SHA, Facility cannot reliably determine whether a run was based on the current tree, reconstruct the exact delivered range, or provide durable provenance in receipts.

The runner already calculates these commits during workspace preparation and delivery; this change persists them instead of discarding them.

This is additive and backward-compatible:

- Existing database rows remain valid.
- Old receipts remain valid because `base_sha` is optional.
- No historical backfill is attempted.
- Gate 1 UI and stale-base enforcement are intentionally outside this PR.

Refs #147.

## Verification

Verified:

- Workspace SHA validation and lowercase normalization
- Authenticated first-write and idempotent replay behavior
- Conflicting replay rejection
- Resume over an advanced base records the restored workspace `HEAD`
- Delivery base propagation and database persistence
- Receipt fallback, legacy compatibility, and integrity hashing
- Database migration
- OpenAPI and generated SDK consistency
- Relevant builds and typechecks
- Complete runner package suite: 169 tests passed
- SDK suite: 23 tests passed
- Biome and `git diff --check`

`pnpm verify` is not marked as passing locally. WSL verification encountered timing-sensitive failures in unchanged runner process-deadline and assistant cleanup tests. The affected runner tests pass independently, and no provenance test failed.

- [x] Behaviour verified beyond the focused suite
- [x] Generated API documentation is updated; no UI change


UsamaIslam:
Nice catch on splitting `workspace_base_sha` from delivery `base_sha` — that's the right shape, since the two answer different questions (what the agent read vs. what got published). One thing I couldn't tell from the PR description: does anything compare `workspace_base_sha` against the live branch HEAD at plan-acceptance time (before `/builder` is dispatched), or does this land the data without yet wiring it into the Gate 1 approval UI? If it's just the former for now, might be worth a follow-up issue so "record" and "surface to the approver" don't silently get conflated as the same fix.

jungle-lethanh:
Good question. This PR only lands the durable provenance data.
It does not currently resolve the live branch HEAD or compare it with `workspaceBaseSha` during plan acceptance, and it does not change the Gate 1 UI or `/builder` dispatch behavior.
The intended follow-up would use the plan-producing run’s `workspaceBaseSha` as the recorded base, resolve the current target-branch HEAD at approval time, and surface the drift to the approver. Whether that should warn or block dispatch is a separate policy/UX decision.

I kept capture and persistence separate from Gate 1 behavior so the provenance contract can be reviewed independently. I used `Refs #147` rather than closing the issue, and I’ll update the PR description to make this scope boundary more explicit. We can keep the Gate 1 work under #147 or track it in a focused follow-up issue, depending on maintainer preference.

adrian-lorenzo (approved):

jungle-lethanh:
Thank you for the review and approval! I appreciate the opportunity to contribute.

Once the provenance PR lands, I’d be interested in taking the remaining Gate 1 slice.
My proposed first scope is to compare the architect run’s workspaceBaseSha with the live target-branch HEAD, then show an advisory at platform plan approval with the current SHA and the paths changed since planning. I would recheck at approval time and avoid automatically blocking dispatch in the first slice.

I would not infer overlap by parsing the plan Markdown; the UI would present deterministic changed-path evidence for the human decision.
Should this remain under #147, and would you prefer the first slice to cover only the platform approval UI, or also define behavior for GitHub /builder acceptance?

adrian-lorenzo:
Thanks again for the provenance work, it was incorporated into #205 with your authorship preserved.

#205 also completed the remaining Gate 1 work under #147 for both the platform approval flow and GitHub `/builder`. It uses fail-closed base-SHA and issue-revision revalidation, so there is no separate advisory-only slice left to take here.

I’ve updated this branch to current `main` and resolved the conflicts. The PR now has no remaining diff because its work is already present on `main` through #205.

Issue #28, opened 2026-07-22
checks_not_configured surfaces after the paid run — preflight config validation + human-readable errors


javiertoledo:
Found operating a project from the UI. A `builder` was dispatched from the web on a project connected **without kickstart** — so `settings.check_cmds` was `[]`. The run executed fully (4m, $3.98, engine exit 0, progress gate green) and only THEN the delivery gate failed with `checks_not_configured`: no branch, no PR, work discarded. Run: `run_019f871d9db779aebb317f015b7923fa`.

The gate itself is correct (a builder without acceptance checks can't prove anything — fail closed is right). The problem is **when** and **how** it surfaces:

1. **Detect at dispatch time, not after the paid run.** `checkCmds` resolution happens in the orchestrator before the sandbox launches (`resolveCheckCmds(profile, project.settings)`); if a delivery-mode agent (`requiresDelivery`) is dispatched with zero check commands, the trigger endpoint / UI button can refuse (or hard-warn) immediately: "This project has no acceptance checks configured — a builder run cannot deliver. Configure them in Settings." Cost of the miss today: the full agent run.
2. **Human-readable failure.** The session page shows raw `{"code":"checks_not_configured"}` with no explanation and no link to the fix. Error codes need a copy layer: what it means, why the work wasn't shipped, where to fix it (project Settings), and a retry affordance once fixed.
3. **Connect-without-kickstart should ask.** The no-kickstart path (observe-first, theam/facility#26) skips check detection entirely; connecting a repo could prompt for check commands (or infer: `npm test` when package.json has a test script) so the first builder doesn't trip on this landmine.

## Acceptance
- [ ] Dispatching a delivery-mode agent on a checks-less project is blocked (or explicitly confirmed) before any sandbox is provisioned.
- [ ] `checks_not_configured` renders as prose with a link to project Settings.
- [ ] Repo connect flow captures/infers check commands.

PR #168, opened 2026-08-16
fix(api): preflight missing repository before builder dispatch

elirethDev:
Related to #28.

## What changes

A builder run without a repository cannot create a branch or pull request. Before this change, it could provision the sandbox and spend the run budget before the runner reported `delivery_repo_not_configured`.

- `dispatchRun` now fails a builder without a repository before the atomic claim, spend-capable key creation, or sandbox provisioning, using `delivery_repo_not_configured`.
- Connected GitHub repository triggers no longer run an API-side check-command preflight. Under the current contract, GitHub CI owns acceptance for those deliveries.
- The session page renders `delivery_repo_not_configured` as a human explanation with a link to project Settings. Unknown errors still render raw, and the existing `checks_not_configured` presentation remains available as a runner backstop.

The connected-repository/no-kickstart scenario described in #28 appears to have been superseded by the current GitHub-CI-owned acceptance model. The repo-connect prompt from the original issue remains outside this change.

## Files

| File | Change |
|------|--------|
| `services/api/src/sandbox/delivery-gate.ts` | Delivery-mode and repository preflight predicate |
| `services/api/src/sandbox/orchestrator.ts` | Fail repo-less builders before claim, key creation, or provisioning |
| `services/api/src/routes/v1/github.ts`, `services/api/src/github/router.ts` | Remove the obsolete check-command trigger preflights |
| `apps/web/components/run/run-error.ts` | Human-readable `delivery_repo_not_configured` presentation |
| `services/api/test/*`, `apps/web/test/run-error.test.ts` | Regression coverage for repo-less builders and connected GitHub repositories |

## Verification

- API focused tests: 90 passed, 1 timing-based timeout in `github-platform-lane.test.ts` (`coalesces a CI notification burst`), 0 skipped; the database-backed suites ran with Docker/Postgres healthy
- Web error presentation test: 8/8 passed
- `pnpm --filter @facility/api typecheck` and `pnpm --filter @facility/web typecheck`: pass
- `pnpm exec biome check` on the changed files: clean
- `git diff HEAD^ HEAD --check`: clean
- Full `pnpm verify` remains blocked at its first Lint step by the existing Windows `scripts/verify.mjs` `spawnSync("pnpm")` issue (`ENOENT`); the full acceptance suite remains for CI.


services/api/src/sandbox/acceptance-gate.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!bundle.repo.cloneUrl?.startsWith("https://github.com/")) return false;
+  return requiresDelivery(bundle.mode) || repairRepositoryMode(bundle.mode);
+}
+
+// The runner's acceptance gate, evaluated at dispatch time: a delivery-mode
+// agent may only run when acceptance is configured or the repository's own CI
+// owns acceptance.
+export function checksConfiguredForDispatch(bundle: AcceptanceBundle): boolean {
+  return (
+    githubCiOwnsAcceptance(bundle) || !requiresDelivery(bundle.mode) || bundle.checkCmds.length > 0
```
adrian-lorenzo: This doesn’t seem to catch the case from #28 anymore. `checksConfiguredForRepo()` gives every connected repo a GitHub URL, so this always passes for a builder even when `checkCmds` is empty. The only builder it blocks has no repo; asking that user to configure checks is misleading, because the run will then start and fail with `delivery_repo_not_configured`. Could we either preflight the missing repo too, or re-scope this now that GitHub CI owns acceptance?

adrian-lorenzo (changes requested):

services/api/src/sandbox/acceptance-gate.ts
elirethDev: Mmm, yeah, you're right. Looking at the current flow, I think the right thing to do is:

- replace the current API-side checks preflight with an early `delivery_repo_not_configured` guard for builder runs without a repository, before creating keys or provisioning a sandbox;
- remove the `checksConfiguredForRepo()` preflight from connected GitHub repository triggers;
- add human-readable UI copy and regression tests for that failure mode;
- update the PR description to make it clear that GitHub CI owns acceptance for connected repositories.

That way, connected repositories follow the current GitHub CI acceptance model, while builder runs without a repository fail early instead of spending resources on a run that cannot deliver.

elirethDev: Implemented in `19d5d88`.

Connected GitHub repositories no longer run the check-command preflight. Builder runs without a repository now fail before key creation or sandbox provisioning with `delivery_repo_not_configured`, with matching UI copy and regression coverage. I also updated the PR title and description to reflect the current GitHub CI acceptance model.

Focused checks pass: 32 API tests, 8 web tests, both typechecks, Biome, and diff whitespace checks.

elirethDev: I reran the database-backed suites with Docker and Postgres healthy. The result was 90 passed, 1 timing-based timeout in `coalesces a CI notification burst`, and no skipped tests.

I also confirmed that `pnpm verify` still stops at its first Lint step on Windows with `scripts/verify.mjs` calling `spawnSync("pnpm")` and receiving `ENOENT`.

adrian-lorenzo (changes requested):
Thanks for following up on the earlier review!

The corrected repository preflight works on this branch, but `dispatchRun` now conflicts with the two-phase Builder-plan gate on `main`. Please preserve that gate and its atomic execution snapshot, then apply the repository check after `dispatchSnapshot` is created but before any keys or sandbox resources exist. Add an integration regression covering a plan-required Builder without a repository so conflict resolution cannot accidentally weaken Gate 1.

There is also a small UI mismatch: the new repository error links to Settings with the label “configure acceptance checks.” Please make the action label describe connecting a repository.

Once those integration details are resolved, this should be ready to approve.

elirethDev:
> Thanks for following up on the earlier review!
> 
> The corrected repository preflight works on this branch, but `dispatchRun` now conflicts with the two-phase Builder-plan gate on `main`. Please preserve that gate and its atomic execution snapshot, then apply the repository check after `dispatchSnapshot` is created but before any keys or sandbox resources exist. Add an integration regression covering a plan-required Builder without a repository so conflict resolution cannot accidentally weaken Gate 1.
> 
> There is also a small UI mismatch: the new repository error links to Settings with the label “configure acceptance checks.” Please make the action label describe connecting a repository.
> 
> Once those integration details are resolved, this should be ready to approve.

For some reason, my email system sent this reply to the spam folder. I'll take care of it today so you can proceed with the merge. Best regards!

javiertoledo:

Both ends of it are gone: `sandbox/delivery-gate.ts` and `sandbox/orchestrator.ts` on the API side, and `apps/web/components/run/` on the web side. None of your 9 files are on main.

Failing early with a clear error instead of dispatching into a missing repository still applies to the new dispatch path — `services/api/src/turns/dispatcher.ts` is where that decision happens now, and `workspaces/runtime.ts` is what it provisions.

Thanks for the work, and for staying with it through the review rounds.

PR #169, opened 2026-08-16
docs: document the repository settings behind the human gates

### LINKED ISSUE #48: Document the repository settings that back the human gates
Facility's core claim is that agents cannot approve their own work, cannot merge, and cannot push to protected branches. Some of that is enforced by Facility; some of it depends on how the GitHub repository is configured, and the documentation does not currently say which is which.

That gap is easy to hit: a repository can run the whole loop, look correct, and still allow an automated actor to satisfy a human gate — for example when *Allow GitHub Actions to create and approve pull requests* is left enabled, or when the default branch has no protection at all.

## Scope

- A short section — likely in the security reference, linked from the kickstart and existing-repo guides — listing the repository settings a Facility-governed repository should have, and stating for each whether Facility enforces it or the setting does.
- Cover at least: Actions approving pull requests, default `GITHUB_TOKEN` permissions, default-branch protection and required checks, and who may merge.
- `facility doctor --github` already checks parts of the GitHub side. Where a setting is checkable through the API, say so, and note the ones worth adding to `doctor` as a follow-up.

## Acceptance

- [ ] A reader can tell, per setting, whether skipping it weakens a gate the product advertises.
- [ ] The guides point at it before the first agent run, not after.

alejandro-llanes:
## What changes

`apps/docs/docs/reference/security.md` gains a **Repository settings that back the gates**
section, placed immediately after *The invariants that never move* — the paragraph it
qualifies. For each setting it states the required value, **whether Facility or the setting
enforces it**, and which advertised gate a skip breaks.

It covers the four areas the issue asks for — Actions approving pull requests, the default
`GITHUB_TOKEN` scope, default-branch protection and required checks, and who may merge or
bypass — plus stale-approval dismissal, which matters here specifically because the
address-review agent pushes commits to a pull request branch after a human has reviewed it.

Two subsections follow: which settings are readable through the GitHub API (endpoint and
field), and what `facility doctor --github` does **not** check yet, recorded as a follow-up
rather than implemented here.

Three guides link to it before the first agent run:

| file | where |
|---|---|
| `guides/kickstart.md` | step 5, and again immediately before the `/architect` paragraph |
| `guides/existing-repo.md` | Step 1 — import |
| `reference/cli.md` | where `doctor --run-guards --github` is described, to state the boundary of what it checks |

Documentation only. No code, no behaviour change.

## Why

Closes #48.

The gap is easy to confirm: grepping the repository for `approve pull requests`,
`can_approve`, `approving_review_count`, `enforce_admins` or `restrictions` returns nothing.
The setting *Allow GitHub Actions to create and approve pull requests* is not mentioned
anywhere, and the existing guidance ("protect the default branch, require a human review")
names no setting and assigns no ownership.

That matters because the invariants lean on repository configuration more than the docs
admit:

- `facility-review.yml` states it *"never approves or merges"* — but that is a prompt, and
  the job runs with `pull-requests: write`. The hard stop is
  `can_approve_pull_request_reviews: false`.
- *Never push to protected branches* is vacuous on an unprotected default branch, and
  `facility-address-review.yml` and `facility-doctor.yml` hold `contents: write`.
- The GitHub App installation holds `Contents: Read and write`, so a bypass entry on
  required pull requests converts that into merge authority.

So a repository can run the whole loop, look correct, and still let an automated actor
satisfy a human gate — which is exactly what the issue describes.

## Verification

Every API claim in the new section was probed against the live GitHub API rather than
written from memory:

| probe | result |
|---|---|
| `GET /repos/{owner}/{repo}/actions/permissions/workflow` | returns both settings in one call: `{"default_workflow_permissions":"read","can_approve_pull_request_reviews":false}` |
| `GET /orgs/{org}/actions/permissions/workflow` | `403` without org admin and the `admin:org` scope |
| `GET /repos/{owner}/{repo}/branches/{branch}/protection` | `404 "Branch not protected"` as an admin on an unprotected branch, but `404 "Not Found"` without admin — two different meanings, same status code |
| `GET /repos/{owner}/{repo}/rules/branches/{branch}` | answers **without** admin, so it works where the protection endpoint cannot |

Those last two are why the `doctor` follow-up list is specific: the command currently treats
any non-zero exit as a failure, so a branch protected by a ruleset and a caller who simply
lacks admin both report the same way as a genuinely unprotected branch — and the protection
response it already fetches is never parsed.

Commands run:

```
node guards/run.mjs                  # actions-pinned, markdown-links — 2 passed, 0 failed
pnpm --filter @facility/docs test    # 12 passed, 0 failed (includes the Pages build)
pnpm verify                          # see below
```

The docs build runs under `onBrokenLinks: "throw"`, so the four new relative links are
verified rather than assumed. Because the guides link to an anchor and `onBrokenAnchors`
only warns, I checked the built output directly instead of trusting the slug rule:
`id="repository-settings-that-back-the-gates"` is present in `reference/security/index.html`,
and `guides/kickstart`, `guides/existing-repo` and `reference/cli` all emit links pointing at
it.

### On `pnpm verify`

It completed Lint, Typecheck, the clean cache-disabled workspace build, isolated Postgres,
both test-database recreations, and the critical integration tests, then exited `1` in the
final stage on a single test:

```
FAIL runner/test/docker-proxy.test.ts > restricted Docker API
     > forwards a Docker exec upgrade body before waiting for 101
Error: Test timed out in 5000ms.        (then afterEach: Hook timed out in 10000ms)

Test Files  1 failed | 11 passed (12)
     Tests  1 failed | 168 passed (169)
```

**This is pre-existing and unrelated to this pull request.** I confirmed it rather than
assuming: checking out unmodified `main` at `ae68401` and running that file alone reproduces
the same failure, same test, same timeout —

```
× forwards a Docker exec upgrade body before waiting for 101  15011ms
Test Files  1 failed (1)
     Tests  1 failed | 24 passed (25)
```

It looks environment-dependent (a Unix-socket upgrade handshake that never receives its
`101`), so it may well be green in CI — flagging it only so the failing run below isn't
mistaken for something this branch introduced. Nothing in this diff is reachable from
`runner/`; the change is four Markdown files under `apps/docs/`.

- [ ] `pnpm verify` passes locally — runs green **except** the pre-existing
      `docker-proxy` timeout documented above, which also fails on clean `main`

### DIFF
--- a/apps/docs/docs/guides/existing-repo.md
+++ b/apps/docs/docs/guides/existing-repo.md
@@ -13,6 +13,12 @@ Connect the repo to a project. The platform detects vendored facility files
 and records the current state as the fingerprint baseline (**adopt**), so
 integrity checking starts from reality, not from an ideal.
 
+An adopted repository has been running agents without Facility's assumptions
+being checked. Verify the
+[repository settings that back the gates](../reference/security#repository-settings-that-back-the-gates)
+now: fingerprinting records what the repository is, not whether its human gates
+can actually hold.
+
 ## Step 2 — money first
 
 Switch the repo's provider secrets to gateway virtual keys
--- a/apps/docs/docs/guides/kickstart.md
+++ b/apps/docs/docs/guides/kickstart.md
@@ -22,6 +22,10 @@ From zero to a working factory:
    `facility/kickstart` and opens a pull request. Manual steps that only you
    can do are in the PR body: create the agent token secret, protect the
    default branch, confirm App permissions, and complete preview configuration.
+   Work through the
+   [repository settings that back the gates](../reference/security#repository-settings-that-back-the-gates)
+   while you are there — several of the human gates are held up by repository
+   configuration rather than by Facility.
 6. **Configure a live PR preview.** Choose a Facility-owned preview or an
    external deployment adapter. For a native preview, provide an immutable
    image, optional command, internal port, readiness path, and TTL. Add
@@ -32,6 +36,11 @@ From zero to a working factory:
 7. **Merge it.** That's Gate 2 muscle memory from day zero. Validate the live
    preview, review the PR, and squash-merge it in GitHub. On merge the
    fingerprint baseline is recorded and the project reports **system ok**.
+
+Confirm the
+[repository settings that back the gates](../reference/security#repository-settings-that-back-the-gates)
+are in place before this point: the gates have to exist before the first agent
+run, not after the first surprise.
 
 Now open an issue and comment `/architect`. The agent's task-specific checklist
 and final plan appear in one comment. Continue entirely from GitHub: comment
--- a/apps/docs/docs/reference/cli.md
+++ b/apps/docs/docs/reference/cli.md
@@ -71,7 +71,10 @@ Actions in the repository lane and by Facility in the platform lane.
 
 Run `facility doctor --run-guards --github` after committing. It checks the
 manifest, generated workflows, configured agent models and authentication,
-preview variables/secrets, deterministic guards, and branch protection.
+preview variables/secrets, deterministic guards, and branch protection. It does
+not check every setting the human gates depend on — see
+[repository settings that back the gates](../reference/security#repository-settings-that-back-the-gates)
+for the full list and for what the command does not yet cover.
 
 ## Platform lane
 
--- a/apps/docs/docs/reference/security.md
+++ b/apps/docs/docs/reference/security.md
@@ -101,4 +101,69 @@ Agents never approve, never merge, never push to protected branches. Every
 outward action carries a named principal. Every merge carries a human
 decision.
 
+## Repository settings that back the gates
+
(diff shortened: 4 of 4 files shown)


adrian-lorenzo (changes requested):
Thanks for putting this together!

The checklist is useful, but the API section cannot currently verify the bypass guarantee it asks operators to enforce. Under classic branch protection, `restrictions` covers who can push; pull-request bypass actors live in `required_pull_request_reviews.bypass_pull_request_allowances`. For rulesets, `/rules/branches/{branch}` returns active rules but not the ruleset’s `bypass_actors`; those require fetching each ruleset, and GitHub only exposes them when the caller has sufficient access.

Please point readers to the correct fields and endpoints, make the result explicitly unknown when permissions hide bypass actors, and clarify that `admin:org` is the classic-token requirement—fine-grained and App tokens use the organization Administration permission.

Once those distinctions are reflected in the guidance, this will be ready to approve.

PR #173, opened 2026-08-17
fix: several Windows-checkout issues found while running the platform's own checks locally

### LINKED ISSUE #2: capataz update: re-sync vendored files (mechanical for machine-owned, AI-proposed merge for edited files)
## Problem

There is no way to pull SDLC improvements into a repo that already ran `init`. When the method evolves (a workflow hardens, a guard is added, the standard gains a rule), every consuming repo is frozen at the version it was installed with. This is the **#1 roadmap item** in the README (`capataz update` — re-sync vendored files, three-way).

The hard part is capataz's own philosophy: **everything is vendored, and `STANDARD.md` is *meant* to be edited.** So `update` cannot overwrite — it has to merge.

## Proposed design — split files by ownership

Capataz files fall into two classes, and they want **different** update strategies:

**1. Machine-owned — regenerated from `.capataz.json`** (workflows, `guards/`, `.claude/hooks/*`, `move-board-status.sh`, `/verify` + `/open-pr`). These are rendered deterministically from the manifest's choices and are *not* meant to be hand-edited. → A deterministic re-render (or plain three-way) is safe and mostly mechanical.

**2. Human-owned — meant to evolve in the repo** (`STANDARD.md`, the managed blocks in `AGENTS.md`/`CLAUDE.md`, `.claude/skills/*`). These carry the team's accumulated opinions. → A mechanical merge will fight the user.

For class 2, rather than a plain `npx` three-way merge, **run an agent/skill (Claude or Codex) that proposes a semantic merge** — it reads the new upstream version, the user's edited version, and the common ancestor, and opens a diff/PR the human reviews and signs. This is exactly capataz's own model: *agents propose, humans sign.* It degrades gracefully where a line-based three-way merge would produce garbage on prose.

## Open questions to resolve

- [ ] **Baseline storage.** `.capataz.json` records the version but not pristine file baselines/hashes. A real three-way needs the common ancestor — either re-fetch the old templates by version from the published package, or store baselines at install time. (Depends on npm publish.)
- [ ] Is `update` a single command that dispatches per-class, or `update` (mechanical) + a separate skill invocation for the prose files?
- [ ] Engine-agnostic: the proposal skill should work under both Claude and Codex (consistent with the multi-engine roadmap).
- [ ] How `add`ed modules participate (their `STANDARD.md` sections + spliced hook rules use the same marker machinery).

## Why it matters

Publish (#1) + update together are what turn capataz from "scaffold once" into "a living SDLC standard we can roll improvements out to every repo." This is the multiplier.

aakashkumarjaiswal0987-ops:
## Summary

While setting up this project locally (no Docker available, so only the Node-only paths: guards, typecheck, and the dev test suite), I found and fixed three related issues that are all invisible in CI because every job in `.github/workflows/ci.yml` runs on `ubuntu-latest`. I also found a fourth issue that I'm reporting but deliberately did not fix — see the note at the end.

### 1. `guards/markdown-links.mjs` — false-positive "missing" links on Windows

The guard resolves the repository root via `git rev-parse --show-toplevel`, which always returns forward-slash paths, even on Windows, then compares it against locally-resolved link targets with `startsWith`. Node's `path.resolve()` / `path.sep` return backslashes on Windows, so the containment check never matches — **every** local Markdown link in the repo (README.md, CONTRIBUTING.md, docs) gets reported as missing, even though the files exist.

Fix: wrap the git output in `resolve()` so it normalizes to the platform's native separators before comparison.

```
$ node guards/run.mjs          # before
✗ markdown-links   (58 false-positive violations)
$ node guards/run.mjs          # after
✓ markdown-links   (0 failures)
```

### 2. No `.gitattributes` — checkout line endings depend on contributor config

This repo has no `.gitattributes`, so whether a checked-out file ends up LF or CRLF depends entirely on a contributor's local `core.autocrlf`. On Windows with the common `core.autocrlf=true` default, this silently introduces CRLF bytes into files the reviewed source never had them in, and breaks anything that hashes or pattern-matches file content byte-for-byte. This is the root cause behind issue #3 below and behind 11 additional (already-fixed-by-normalization) false negatives in `scripts/release.test.mjs` / `scripts/release.integration.test.mjs` that compared regex patterns against generated workflow YAML.

Fix: added `.gitattributes` pinning `* text=auto eol=lf`, so every tracked text file checks out identically regardless of contributor platform.

### 3. `scripts/dependencies-security.test.mjs` — false-negative DoS-regression test

This test verifies the local security patch for `image-size@2.0.2` (rejecting non-advancing ICNS/HEIF/JXL boxes) is applied, by string-interpolating the patch hash from `pnpm-lock.yaml` into a hardcoded `node_modules/.pnpm/image-size@2.0.2_patch_hash=<hash>` path. Before fix #2, `patches/image-size@2.0.2.patch` checked out with CRLF on Windows, so pnpm computed a different content hash for it than the lockfile recorded, and the installed directory didn't match the hardcoded path — throwing `ERR_MODULE_NOT_FOUND` instead of actually checking the patch.

I confirmed the underlying patch **was** correctly applied in the real installed package the whole time (`grep "does not advance" .../dist/fromFile.cjs` finds all three guards) — this was purely a test bug, not a broken patch, but it's a false negative on a security-relevant test, which seemed worth fixing regardless of #2.

Fix: scan `node_modules/.pnpm` for an entry matching `image-size@2.0.2_patch_hash[=_]<hash>` instead of constructing the exact path, so the test checks the actual invariant ("a patched image-size@2.0.2 is installed") without depending on an exact hash/separator.

### 4. `tar` invoked without `--force-local` misparses Windows drive letters as remote hosts

`scripts/release.mjs` (`inspectTarball`, real release logic — not just a test) and `scripts/release.integration.test.mjs`'s fixture both call `tar` with paths from `path.resolve()` / `os.tmpdir()`, which are drive-letter paths on Windows (e.g. `C:\Users\...`). GNU tar (what Git for Windows puts on `PATH`) interprets the colon as a remote-host separator unless told otherwise, and fails with `tar (child): Cannot connect to C: resolve failed` instead of reading the local path. This means running the actual release script — not only its tests — would fail on a Windows machine today.

Fix: pass `--force-local` to both invocations. It's a no-op on Linux/macOS, so no CI behavior changes.

```
$ node --test scripts/release.test.mjs scripts/release.integration.test.mjs   # after gitattributes + this fix
24/31 passing (up from 13/31 before any fix in this PR)
```

### Found, not fixed: `spawnSync("npm", ...)` in `release.mjs` throws ENOENT on Windows

The remaining 7 failing release tests all fail with `spawnSync npm ENOENT`. On Windows, `npm` on `PATH` is `npm.cmd`, and `spawnSync`/`execFileSync` don't resolve `.cmd` shims without `shell: true` (confirmed: `npm.cmd` directly throws `EINVAL` here; `shell: true` works but Node emits a `DEP0190` warning because unescaped shell args are a real risk). This is inside `publishTarball()`, which also builds npm auth config and handles `NODE_AUTH_TOKEN` — I didn't feel comfortable proposing a fix to that code path without your input on the right tradeoff (`shell: true` with careful arg escaping vs. a `cross-spawn`-style dependency vs. something else), so I'm flagging it rather than guessing. Happy to take a pass at it if you'd like a specific direction.

## Verification

Every fix above was verified by toggling it off and back on with `git stash` / `git stash pop` in the same terminal session and re-running the affected check each time, to confirm the before/after wasn't coincidental — not just eyeballing a single run.

## Test plan

- [x] `node guards/run.mjs` — 2/2 passing (was 1/2)
- [x] `node --test scripts/dependencies-security.test.mjs` — 3/3 passing (was 0/3), underlying patch confirmed genuinely applied independent of the test bug
- [x] `node --test scripts/release.test.mjs scripts/release.integration.test.mjs` — 24/31 passing (was 13/31); remaining 7 are the documented npm ENOENT issue, unrelated to this PR's fixes
- [x] `pnpm typecheck` — 16/16 packages, unaffected
- [x] No behavior change on Linux for any of the four fixes (`.gitattributes` normalizes to what Linux already produces; `resolve()` and `--force-local` are no-ops when paths already use `/`)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,18 @@
+# Normalize line endings to LF for every tracked text file, regardless of a
+# contributor's local core.autocrlf setting. Several guards and tests compare
+# file content or hash it (e.g. guards/markdown-links.mjs, patches applied via
+# pnpm, CI-workflow regex assertions in scripts/release*.test.mjs) and break or
+# silently diverge when a checkout introduces CRLF bytes that were not in the
+# reviewed source.
+* text=auto eol=lf
+
+# Explicitly binary — never subject to line-ending normalization or diffing.
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.woff binary
+*.woff2 binary
+*.ttf binary
+*.otf binary
--- a/guards/markdown-links.mjs
+++ b/guards/markdown-links.mjs
@@ -138,9 +138,11 @@ export default {
   name: "markdown-links",
   description: "tracked Markdown files do not link to missing local targets",
   run() {
-    const root = execFileSync("git", ["rev-parse", "--show-toplevel"], {
-      encoding: "utf8",
-    }).trim();
+    const root = resolve(
+      execFileSync("git", ["rev-parse", "--show-toplevel"], {
+        encoding: "utf8",
+      }).trim(),
+    );
     const violations = [];
 
     for (const file of trackedMarkdownFiles(root)) {
--- a/scripts/dependencies-security.test.mjs
+++ b/scripts/dependencies-security.test.mjs
@@ -1,6 +1,6 @@
 import assert from "node:assert/strict";
 import { spawnSync } from "node:child_process";
-import { readFile } from "node:fs/promises";
+import { readdir, readFile } from "node:fs/promises";
 import { dirname, join, resolve } from "node:path";
 import test from "node:test";
 import { fileURLToPath, pathToFileURL } from "node:url";
@@ -11,16 +11,16 @@ async function patchedImageSizeModule() {
   const lockfile = await readFile(join(root, "pnpm-lock.yaml"), "utf8");
   const match = lockfile.match(/^  image-size@2\.0\.2: ([a-f0-9]+)$/m);
   assert.ok(match, "image-size 2.0.2 must remain pinned to the reviewed local patch");
-  return join(
-    root,
-    "node_modules",
-    ".pnpm",
-    `image-size@2.0.2_patch_hash=${match[1]}`,
-    "node_modules",
-    "image-size",
-    "dist",
-    "index.mjs",
+
+  const storeDir = join(root, "node_modules", ".pnpm");
+  const entries = await readdir(storeDir);
+  const patchedEntry = entries.find((entry) => /^image-size@2\.0\.2_patch_hash[=_][a-f0-9]+$/.test(entry));
+  assert.ok(
+    patchedEntry,
+    "expected a patched image-size@2.0.2 entry under node_modules/.pnpm — run pnpm install",
   );
+
+  return join(storeDir, patchedEntry, "node_modules", "image-size", "dist", "index.mjs");
 }
 
 function writeBox(buffer, offset, size, name) {
--- a/scripts/release.integration.test.mjs
+++ b/scripts/release.integration.test.mjs
@@ -32,7 +32,7 @@ async function fixture(t) {
     join(packageDir, "lifecycle.mjs"),
     "import { writeFileSync } from 'node:fs';\nwriteFileSync(process.env.RELEASE_LIFECYCLE_MARKER, 'ran');\n",
   );
-  execFileSync("tar", ["-czf", tarball, "-C", root, "package"]);
+  execFileSync("tar", ["--force-local", "-czf", tarball, "-C", root, "package"]);
 
   t.after(async () => {
     await rm(root, { recursive: true, force: true });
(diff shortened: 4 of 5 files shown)


adrian-lorenzo (changes requested):
Thanks for the contribution!

`--force-local` is specific to GNU tar. macOS’s `bsdtar` rejects it, causing 18 of the 31 release tests to fail. Please avoid the platform-specific flag—for example, run tar from the archive’s directory and pass only its basename—and add coverage for both Windows drive-letter paths and BSD tar.

The `image-size` test now reads the expected patch hash from the lockfile but ignores it when selecting the installed directory. This can pass against a stale or different patched package. Please keep the selected installation tied to the expected hash and add a regression with multiple candidate directories.

The Markdown path normalization has already landed on `main`, so please update the branch and remove the duplicate change. Please also reword “Before fix #2,” which currently causes GitHub to mark the unrelated issue #2 for closure.

Once these are addressed, the remaining Windows portability work should be in good shape.

Issue #110, opened 2026-08-07
Model and show which stories block which


javiertoledo:
> **As** someone looking at a story,
> **I want** to see what it is waiting on and what is waiting on it,
> **so that** I can tell the difference between work that is merely queued and work whose completion releases three other things.

## What is true today

**Facility models no relationship between stories at all.** There is no dependency, blocker, parent or sub-issue concept anywhere in the schema. The only issue-to-issue edge that exists is `gh_pull_requests.closing_issues` (#67), and that one is about *identity* — this PR and this issue are the same unit of work — not about order.

**The word "blocked" is already taken, and means something else.** In Facility it means *a run awaiting a human gate*: `status === "awaiting_human"` (`apps/web/app/(app)/projects/[projectId]/page.tsx:82`, `apps/web/app/(app)/(org)/projects/page.tsx:44`), surfaced as "needs you". Whatever this concept ends up being called, it will appear on the same screens as that one, and the two must not blur.

**The information already exists — as prose.** *"depends on #877"*, *"after #886 lands"*, *"blocked on the data platform"*, written into issue bodies and comments every day. #98 will make those references clickable; it will not make them *known*. Nothing can count them, sort by them, or notice when one is resolved.

**And it is the missing third of the prioritisation question.** The companion issue #109 asks the board to order work by urgency, business value and *what it unblocks*. Facility can approximate the first two from what the Project Owner agent already scores. The third it cannot compute at all, because the edges do not exist.

## What good looks like

- On a story, **both directions are visible and unmissable**: what it is waiting on — with each blocker's current stage — and what is waiting on it.
- On the board, the difference is legible at a glance. A Backlog story with an unfinished blocker is not the same as one that is genuinely pickable, and dispatching an agent at the former is usually wasted money.
- A story that releases several others is *visibly* worth more than one that releases none. That is the input #109 needs, and the reason these two are worth thinking about together.
- Edges stay honest over time: a blocker that gets closed stops blocking without anyone remembering to say so, and a cycle is surfaced rather than silently ignored.
- A human can add and remove an edge directly, whatever else creates them.

## Open questions for whoever picks this up

Left open on purpose.

- **Where does the truth live?** GitHub has its own relationship models — sub-issues, and issue dependencies. #67 argued hard, and correctly, for reading GitHub's own answer rather than maintaining a parallel one that GitHub will not honour. Worth establishing what the App's API actually exposes today before choosing to invent anything.
[...]

PR #176, opened 2026-08-17
feat(pipeline): order active stages by the WSJF judgement Facility already records

### LINKED ISSUE #109: Order the backlog by priority, not by last activity
> **As** someone opening the Stories board,
> **I want** the top of each stage to be the work that matters most right now,
> **so that** *"what do I pick up next"* is answered by the board instead of re-derived by hand every morning.

## What is true today

**The board sorts by last activity, not by age and certainly not by importance.** Within a stage, stories are ordered `ghUpdatedAt DESC`, then number DESC, then repo id (`services/api/src/pipeline.ts:317-323`).

That has a perverse property worth naming, because it is the one that actually bites: *any* webhook bumps `gh_updated_at` — a label, an assignment, a comment. Including Facility's own. Triggering an agent posts an acknowledgement comment and, since #70, assigns the triggering user — so **dispatching work moves that story to the top of its stage**. The board reorders itself in response to your own actions, which is the opposite of a stable place to look.

**Reversing it is not the answer either.** "Oldest first" over a last-activity sort puts the most stale — and most likely dead — stories at the top of Backlog. Sorting by `gh_created_at` instead (it is mirrored, `packages/db/src/schema.ts:238`) is at least honest about what it means, but arrival order is not priority. Neither axis knows anything about the work.

**And here is the part that makes this a bug rather than a missing feature: Facility already forms an opinion about priority, and then throws it away.**

- The Project Owner agent's operating contract requires value fields on every task it derives — value, time-criticality, risk-reduction, effort (`packages/harness/contracts/po-agent.md`, enforced by the artifact chain in `packages/harness/src/chain.ts:60-70`).
- They are stored: `po_tasks.wsjf jsonb` (`packages/db/src/schema.ts:886`).
- The scoring and ranking functions exist and are exported: `wsjfScore` and `rankByWsjf` (`packages/harness/src/wsjf.ts`). **Neither has a single caller outside its own unit test.**
- When a task is accepted, the score is serialised into the GitHub issue body as a JSON block under a `## Value` heading, and the issue is labelled `priority:wsjf` (`services/api/src/executors.ts:1868-1881`) — prose in a body, never read back by anything.

So the three variables named in the request map onto a model the codebase already carries: urgency → time-criticality, business value → value, unblocking others → risk-reduction / enablement. The judgement is being made. It is being written down. Nothing consumes it.

## What good looks like

- The default order of a stage reflects a judgement about importance, and that judgement is **inspectable** — I can see *why* something sits on top, not merely that it does.
- The judgement is the Project Owner agent's to make and keep current, and a human's to override — where an override is a decision with a reason, in the record, not a silent drag.
- Priority survives an issue being touched. Commenting on something does not promote it.
- Unscored stories are handled explicitly, not sorted first-or-last by accident. Today the majority of issues are opened on GitHub by hand and have no score at all — **whatever ships has to be true for those**, not only for tasks the PO agent proposed.
- Re-scoring is visible when it happens. A backlog that silently rearranges itself overnight is worse than one that never moves.

## Open questions for whoever picks this up

Left open on purpose — the shape of the answer is the interesting part.

- **Where does the score live?** Facility-native field, or somewhere GitHub can also see it (label, project field, issue body)? #67 set a precedent for reading GitHub's own answer where GitHub has one — here it may not have one, which makes this a genuine choice rather than a lookup.
- **What re-scores, and when?** Priority decays: a bug that was urgent in June may not be in August. A PO run on a schedule, on signals, on both?
- **Is a re-prioritisation a proposal or a fact?** The PO contract says outright: *"never reprioritize the backlog without recording the Decision that justified it."* Whether that Decision needs to pass through Approvals before the board moves is unresolved, and it is the question with the largest blast radius.
- **Where does effort come from?** The agent's estimate, or history — Facility measures real run cost and duration per story and could know more than the estimator did.
- **Per stage, or one queue?** Is there a single project-wide ranked order that the stages merely partition, or seven independent orderings?
- **How much of this is unusable without the dependency graph?** "Unblocks others" is one of the three variables and Facility cannot currently compute it at all — see the companion issue #110.

## What I am deliberately not proposing

A mechanism. No schema, no scoring formula, no UI. The problem statement and the constraints are the contribution; the design belongs to whoever implements it.

Related: #18, #39. Findings are against `5712bf8`.

vicente-debug:
Closes #109.

## What this does

The board no longer sorts active stages by `gh_updated_at`. It sorts them by the WSJF judgement Facility already records — and then threw away. The issue's own framing ("Facility forms an opinion about priority, and then throws it away") is the design: this PR adds no new opinion, it consumes the existing one.

- `assemblePipelineStories` reads the `## Value` section back out of the mirrored issue body (`gh_issues.body_md`) and scores it with the existing `wsjfScore` — the first caller either exported function has had outside its unit test.
- Active stages sort scored stories first, highest score on top. Unscored stories sit below, in GitHub's own newest-created-first order. **Shipped keeps recency**: it is a log of what left the pipeline, not a queue, and "most recently shipped on top" is the honest order there.
- The score and its components ride `PipelineStorySchema` and `StoryDetailSchema`, and the board renders a `wsjf 3.75` chip whose tooltip is the breakdown (`value 8 · time 5 · risk 2 · effort 4`) — the position is inspectable, not merely asserted.
- The serialisation the PO-task executor used to build inline moves into `@facility/harness` (`wsjfValueSection`) next to its new inverse (`parseWsjfValueSection`), so the round trip through the GitHub mirror is pinned by one pair of functions and one pair of tests.

## Where the score lives — the first open question, answered

In the issue body on GitHub, where `executeTaskCreation` already writes it. That choice does the most work per line of code:

- **No second source of truth.** `#101`'s reasoning about assignees applies verbatim: a Facility-native priority field would have no writer, no sync path and no reconciliation story. The body block has all three today — written on task acceptance, mirrored on every webhook and sync, echoed back within seconds of an edit.
- **The human override falls out for free, with a record.** Editing the JSON block on GitHub is an attributed, timestamped issue edit — it lands in the issue history and in Facility's mirror without any new machinery. That is "an override is a decision with a reason, in the record", minus a bespoke approvals flow.
- **Hand-opened issues can opt in** by adding the same `## Value` block the PO agent writes. Without one they are simply unscored, which is handled explicitly, not accidentally: always below scored stories, in stable creation order, with the chip's absence marking the boundary.

Malformed blocks (bad JSON, missing components, `effort <= 0`, a fence that actually belongs to the next section) parse to *unscored*, never to an error — the majority of issues are hand-written, and a parser that can be angered by prose would be a liability.

## What good looks like, checked against the issue

- *"Priority survives an issue being touched."* Nothing in the active-stage comparator reads `gh_updated_at` any more. Triggering an agent — the acknowledgement comment plus the #70 self-assignment that made the board reorder in response to your own actions — now moves nothing.
- *"The judgement is inspectable."* The chip and its breakdown, plus the block itself in the story body.
- *"Unscored stories are handled explicitly."* Deliberately below scored ones, newest-created first; tested.
- *"Re-scoring is visible when it happens."* A re-score is an issue edit: it shows up in the issue history, the timeline's `issue_updated`, and the mirror — not as a silent overnight rearrangement.

## Deliberately left open

The issue's larger questions — PO re-scoring cadence, decay, whether a re-prioritisation must pass through Approvals, effort informed by measured run cost, the #110 dependency graph — are policy, and the issue itself says the design belongs to whoever implements it. This PR keeps the mechanism small enough that any of those answers can be built on top of it: they all reduce to "who is allowed to edit the `## Value` block, and when", because after this change the block is finally load-bearing.

## Tests

- `packages/harness`: round-trip through `wsjfValueSection`/`parseWsjfValueSection`, plus the malformed-body taxonomy.
- `services/api`: a pipeline test where a stale scored story outranks a freshly-commented scored one, unscored stories keep creation order despite activity bumps, and a malformed block lands unscored — end-to-end through `assemblePipelineStories` → `classifyPipeline`.
- OpenAPI + SDK regenerated (`pnpm --filter @facility/api openapi`); full typecheck and the harness/api/web suites pass locally. The one local failure, `ai-identity.test.ts`'s vendor-asset hash, fails identically on a pristine `main` under a Windows CRLF checkout and is what #173 is about.

https://claude.ai/code/session_018XarNuJCzYsJupYqMDUvaG


adrian-lorenzo (changes requested):
Thanks for the contribution!

The `## Value` block is untrusted GitHub content, but the parser currently treats any finite numbers as authoritative. It accepts values the canonical WSJF schema rejects, and values such as `1e308` produce an infinite score that the response schema rejects, allowing one issue to make the Stories endpoint return 500. Any issue author can also promote their own story by editing its body. Please share the canonical validation, ensure the computed score remains finite, bind ranked scores to trusted provenance, and add unit and API-level regressions for negative values, overflow or tiny effort, and forged blocks.

The score breakdown is only exposed through a `title` on a non-focusable element, so it cannot be inspected reliably with a keyboard or touch device. Please make that detail accessible and cover the interaction.

vicente-debug:
Thanks for the thorough review. The overflow-to-500 path and the self-promotion vector were both real.

Addressed in 6e24dcd:

- **Canonical validation is now shared.** `WsjfSchema` moved into `wsjf.ts` and is the same schema task frontmatter uses; the parser and the new `validateWsjf` reject anything non-canonical.
- **Scores are guaranteed finite.** `wsjfScore` refuses non-finite results (a subnormal effort still divides to Infinity even with canonical components), and untrusted paths turn that into "unscored", never an error.
- **Ranks bind to trusted provenance.** Assembly no longer reads the issue body at all: each story's judgement comes from the `po_tasks` row Facility wrote when it mirrored the issue, located through that row's own `gh` reference. Editing the body can neither promote a story nor break the endpoint.
- **The chip is now a native `details/summary` disclosure**, reachable by keyboard, touch, and screen reader, with tests replacing the hover-only `title`.

Regressions cover negative and fractional components, overflow and subnormal effort, and forged blocks at unit and API level, including an issue whose forged block claims 1e308 leaving `/pipeline` at 200.

adrian-lorenzo (approved):
Thanks for the contribution! Thoughtful work throughout.

PR #177, opened 2026-08-18
fix: allow empty strings for optional config variables to prevent crash

### LINKED ISSUE #158: Fresh `pnpm dev` crashes the API and worker: blank VERCEL_* lines in the generated .env fail config validation

webdevsps:
## What changes

By removing `.min(1)` from optional env variables, Zod correctly accepts empty strings instead of throwing a validation error. This allows downstream code to handle them gracefully as unconfigured.

## Why

Fixes #158. 
When `.env` contains blank entries for optional variables (like `VERCEL_TOKEN`), Zod strictly rejects the empty string and crashes the API server on boot.

## Verification

- [x] Behaviour verified beyond the test suite (verified API server boots successfully with empty VERCEL_* env vars)


adrian-lorenzo:
Thanks for the contribution!

This is already covered by #159, which treats blank values as unset and includes regression tests. This version keeps them as empty strings, so a blank `VERCEL_OIDC_TOKEN` can override a valid `VERCEL_TOKEN`. I’m closing this one in favor of #159.

PR #184, opened 2026-08-18
fix(dev): start pnpm through a shell on Windows

### LINKED ISSUE #182: pnpm dev fails on Windows with spawn EINVAL
## Summary

`pnpm dev` cannot start the stack on Windows. It fails immediately with `Error: spawn EINVAL` before Postgres, the API, the gateway or the web app come up.

## Environment

- Windows 10 (10.0.26200), PowerShell
- Node v22.23.2
- pnpm 11.20.0
- `facility` at `ae68401`

## Reproduction

```powershell
pnpm dev
```

## Actual

```
Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at spawn (node:child_process:787:9)
  errno: -4071,
  code: 'EINVAL',
  syscall: 'spawn'
```

## Expected

The development stack starts, as it does on macOS and Linux.

## Root cause

`scripts/dev.mjs` resolves the package manager to a batch file on Windows:

```js
const pnpm = process.platform === "win32" ? "pnpm.cmd" : "pnpm";
```

but `run()` spawns it without a shell:

```js
const child = spawn(command, args, { cwd, env: environment, stdio: "inherit" });
```

Since Node 18.20 / 20.12 (the fix for CVE-2024-27980), `child_process.spawn` refuses to launch a `.cmd` or `.bat` file unless `shell` is set, and raises `EINVAL`.

Every `pnpm` invocation in the script is affected: `pnpm install`, the shared-package build, `@facility/db migrate`, `@facility/db seed`, and `pnpm run dev:services`. The first one aborts the run.

A secondary effect: `spawn` throws **synchronously** here, so the `child.once("error", ...)` handler in `run()` never runs and the friendly `Could not start <label>` hint is replaced by a raw stack trace, which makes the failure harder to diagnose than it needs to be.

Minimal confirmation on the same machine:

```
pnpm.cmd no-shell (current)  : SYNC THROW code=EINVAL
pnpm.cmd shell:true          : exited code=0
docker   no-shell            : exited code=0
```

`docker` is unaffected because it is a real executable, so the fix should be narrow rather than applying a shell to every spawn.

## Suggested fix

Set `shell` only for `.cmd` and `.bat` commands on `win32`, leaving real executables and POSIX platforms with their current argument handling.

Happy to open a PR for this.

mianawais78:
## Summary

Fixes `pnpm dev` on Windows, where it aborted with `Error: spawn EINVAL` before any service started.

`scripts/dev.mjs` resolves the package manager to `pnpm.cmd` on `win32`, but `run()` spawned it without a shell. Since the fix for CVE-2024-27980 (Node 18.20 / 20.12), `child_process.spawn` refuses to launch a `.cmd` or `.bat` file unless `shell` is set. Every `pnpm` step in the script was affected — install, the shared-package build, `migrate`, `seed` and `dev:services` — and the first one ended the run.

The throw is also **synchronous**, so the existing `child.once("error", ...)` handler never ran and the friendly `Could not start <label>` hint was replaced by a raw stack trace.

`shell` is applied only to `.cmd` and `.bat` commands on `win32`, so real executables such as `docker` keep their current argument handling and POSIX platforms are unchanged. The predicate is exported so the behavior is testable without spawning processes.

Closes #182

## Test plan

- [x] `node --test scripts/dev.test.mjs` — new `usesWindowsCmdShell` assertions cover `.cmd` and `.bat` on `win32`, plain commands on `win32`, and `.cmd` on `linux`
- [x] Verified on Windows 10 / Node v22.23.2 that spawning `pnpm.cmd` without a shell throws `EINVAL` synchronously, and succeeds with `shell: true`
- [x] Verified `docker` still spawns without a shell, so the compose steps are unaffected
- [x] `pnpm dev` now brings up Postgres, MinIO, API (`:4400`), worker, gateway (`:4410`), web (`:3400`) and docs (`:3500`) on Windows
- [ ] Confirm on macOS/Linux that `usesWindowsCmdShell` returns `false` and behavior is byte-for-byte unchanged

### Note for reviewers

`scripts/dev.test.mjs` has one failure on Windows that is **pre-existing and unrelated** to this change: the test asserting the generated `.env` mode is `0o600` sees `0o666` (`expected: 384, actual: 438`), because Windows does not honor POSIX file modes. That assertion is present on `main`. I left it alone to keep this PR focused, but I'm happy to address it here or in a separate issue if you prefer.

### DIFF
--- a/scripts/dev.mjs
+++ b/scripts/dev.mjs
@@ -185,13 +185,22 @@ export function developmentServiceEnvironment(prepared, base = process.env) {
   return environment;
 }
 
+export function usesWindowsCmdShell(command, platform = process.platform) {
+  return platform === "win32" && /\.(cmd|bat)$/i.test(command);
+}
+
 export function run(
   command,
   args,
   { cwd = repoRoot, label = [command, ...args].join(" "), environment = process.env } = {},
 ) {
   return new Promise((resolve, reject) => {
-    const child = spawn(command, args, { cwd, env: environment, stdio: "inherit" });
+    const child = spawn(command, args, {
+      cwd,
+      env: environment,
+      stdio: "inherit",
+      shell: usesWindowsCmdShell(command),
+    });
     child.once("error", (error) => {
       const hint =
         error.code === "ENOENT"
--- a/scripts/dev.test.mjs
+++ b/scripts/dev.test.mjs
@@ -10,6 +10,7 @@ import {
   prepareDevEnv,
   setEnvIfBlank,
   usesLocalStorage,
+  usesWindowsCmdShell,
 } from "./dev.mjs";
 
 const example = [
@@ -279,4 +280,12 @@ test("an exported development origin overrides the env file, and a blank one is 
     environment: { FACILITY_DEV_ORIGINS: "   " },
   });
   assert.equal(blank.devOrigins, "from-file.example.dev");
+});
+
+test("Windows .cmd/.bat spawns use a shell so Node does not throw EINVAL", () => {
+  assert.equal(usesWindowsCmdShell("pnpm.cmd", "win32"), true);
+  assert.equal(usesWindowsCmdShell("pnpm.bat", "win32"), true);
+  assert.equal(usesWindowsCmdShell("pnpm", "win32"), false);
+  assert.equal(usesWindowsCmdShell("docker", "win32"), false);
+  assert.equal(usesWindowsCmdShell("pnpm.cmd", "linux"), false);
 });


mianawais78:
### Reproduction detail

Verified on Windows 10 (10.0.26200), Node v22.23.2, pnpm 11.20.0, from `ae68401`.

Spawning the three relevant command shapes directly:

```
pnpm.cmd no-shell (current)  : SYNC THROW code=EINVAL
pnpm.cmd shell:true (fix)    : exited code=0
docker   no-shell            : exited code=0
```

Two details worth noting for review:

1. **The throw is synchronous**, not an `error` event. In `run()` the `spawn` call sits inside the promise executor, so `child.once("error", ...)` is never reached and the `Could not find <command>` / `Could not start <label>` messaging is bypassed entirely. What a Windows user actually sees is a bare `Error: spawn EINVAL` with an internal Node stack:

   ```
   Error: spawn EINVAL
       at ChildProcess.spawn (node:internal/child_process:420:11)
       at spawn (node:child_process:787:9)
     errno: -4071, code: 'EINVAL', syscall: 'spawn'
   ```

2. **`docker` must stay shell-free.** It is a real `.exe` and spawns fine today, so widening `shell` to every command would change quoting behavior for the compose invocations for no benefit. That is why the predicate is narrowed to `.cmd` / `.bat` on `win32` rather than keyed on platform alone.

After the change, `pnpm dev` completes its full sequence on Windows and the stack answers on `:3400`, `:4400`, `:4410` and `:3500`, with `/health` reporting `{"ok":true,"version":"0.3.0","db":"ok"}`.

jungle-lethanh:
I reproduced a related Windows child-process problem in scripts/verify.mjs: pnpm verify fails immediately with spawnSync pnpm ENOENT, before lint starts. PR #184 appears scoped to scripts/dev.mjs, and no active PR currently touches verify.mjs. Would a separate focused issue and PR for the verification command be welcome?

mianawais78:
@adrian-lorenzo sorry for tagging you in the PR, Can you please connect with me https://www.linkedin.com/in/hafiz-muhammad-awais-hassan-518737191
Just need to discuss an idea I have worked around 70 percent need your help and we push it to yc combinator 2026

adrian-lorenzo:
Thanks for the contribution! The Windows process fix is sound, and the targeted test suite passes.

The first commit subject contains an invisible BOM character before `fix(dev)`, so the repository’s conventional-commit guard rejects it. Please rewrite that subject without the leading character and update the branch.

Once that metadata issue is corrected, we can approve it.

PR #185, opened 2026-08-18
fix(api): treat blank optional env values as unset

### LINKED ISSUE #183: API fails to boot from a copied .env.example: blank VERCEL_* values rejected by config schema
## Summary

Copying `.env.example` to `.env` — the documented first step of the self-host quickstart — produces an instance whose API crashes at startup with a `ZodError`, because the example file ships several optional keys with empty values.

## Environment

- Windows 10 (10.0.26200), Node v22.23.2, zod 4.4.3
- `facility` at `ae68401`

## Reproduction

1. `cp .env.example .env` (or let `pnpm dev` create it)
2. Fill in the required values
3. Start the stack

## Actual

The API process exits during import:

```
services\api\src\config.ts:187
  const parsed = EnvSchema.parse(env);
                           ^
ZodError: [
  { "code": "too_small", "minimum": 1, "path": [ "VERCEL_TOKEN" ],
    "message": "Too small: expected string to have >=1 characters" },
  { ... "path": [ "VERCEL_OIDC_TOKEN" ] ... },
  { ... "path": [ "VERCEL_TEAM_ID" ] ... },
  { ... "path": [ "VERCEL_PROJECT_ID" ] ... }
]
    at readConfig (services\api\src\config.ts:187:28)
    at <anonymous> (services\api\src\dev.ts:4:16)
```

## Expected

Optional credentials that are present but blank are treated as unset, and the API boots.

## Root cause

`.env.example` declares these keys with no value (lines 44-47):

```dotenv
VERCEL_TOKEN=
VERCEL_OIDC_TOKEN=
VERCEL_TEAM_ID=
VERCEL_PROJECT_ID=
```

dotenv loads a bare `KEY=` as an empty string rather than omitting the key, which is easy to confirm:

```
VERCEL_TOKEN         present=true value=""
VERCEL_OIDC_TOKEN    present=true value=""
VERCEL_TEAM_ID       present=true value=""
VERCEL_PROJECT_ID    present=true value=""
```

The schema marks the fields optional but validates them with `min(1)`, so `""` counts as present and fails:

```ts
VERCEL_TOKEN: z.string().trim().min(1).optional(),
```

```
z.string().trim().min(1).optional() | absent -> ok | "" -> FAIL (too_small) | "tok" -> ok
```

Because `readConfig` uses `EnvSchema.parse` rather than `safeParse`, this is fatal rather than a warning.

`OptionalUrl` in the same file already normalizes blank strings to `undefined`, so the intended behavior is clear and the affected keys are simply missing that treatment. The same pattern applies to `FACILITY_AWS_CODEBUILD_PROJECT`, `FACILITY_AWS_CODEBUILD_CACHE_BASE_LOCATION` and `PACKAGE_REGISTRY_TOKEN`.

## Suggested fix

Normalize blank strings to `undefined` before validation for these optional fields, so an unset credential stays unset while a whitespace-only value is still rejected rather than silently accepted.

Happy to open a PR for this.

mianawais78:
## Summary

Fixes the API crashing at startup after copying `.env.example` to `.env`, which is the documented first step of the self-host quickstart.

`.env.example` ships `VERCEL_TOKEN`, `VERCEL_OIDC_TOKEN`, `VERCEL_TEAM_ID` and `VERCEL_PROJECT_ID` with no value. dotenv loads a bare `KEY=` as an empty string rather than omitting the key, and the schema marked those fields optional while validating them with `min(1)` — so `""` counted as present and failed. Since `readConfig` uses `EnvSchema.parse` rather than `safeParse`, the result was fatal: the API exited during import with a `ZodError` naming all four keys.

`OptionalUrl` in the same file already normalized blank strings to `undefined`, so the intended behavior was established and these fields were simply missing it. This extracts that normalization as `blankToUndefined`, reuses it for `OptionalUrl`, and adds `OptionalNonEmptyString` for the affected optional credentials, including `FACILITY_AWS_CODEBUILD_PROJECT`, `FACILITY_AWS_CODEBUILD_CACHE_BASE_LOCATION` and `PACKAGE_REGISTRY_TOKEN`.

A whitespace-only value is still rejected rather than silently accepted, so a genuinely malformed credential does not slip through as "unset".

Closes #183

## Test plan

- [x] `pnpm --filter @facility/api exec vitest run test/config.test.ts` — 15 passed, including a new case asserting that blank Vercel credentials from a copied `.env.example` resolve to `undefined`
- [x] Confirmed the previous schema rejected `""` (`too_small`) while accepting `undefined` and a real value, and that the new schema accepts `""` as unset
- [x] Confirmed dotenv reports all four keys as `present=true value=""` when parsing the committed `.env.example`
- [x] API boots from a copied `.env.example` and reports `{"ok":true,"version":"0.3.0","db":"ok"}` on `/health`
- [ ] Confirm no deployment relied on a blank `VERCEL_*` value being a validation error

### Scope note

This deliberately does not change `.env.example`. Removing the blank keys would hide the same class of failure for anyone who writes `KEY=` by hand, and the schema is the correct place to decide that "present but empty" means "unset". Happy to also drop the blank lines from the example file if you would rather have both.

### DIFF
--- a/services/api/src/config.ts
+++ b/services/api/src/config.ts
@@ -8,9 +8,14 @@ import type { AppConfig } from "./types.js";
 const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "../../..");
 loadDotenv({ path: join(repoRoot, ".env"), quiet: true });
 
-const OptionalUrl = z.preprocess(
-  (value) => (typeof value === "string" && value.trim() === "" ? undefined : value),
-  z.string().url().optional(),
+const blankToUndefined = (value: unknown) =>
+  typeof value === "string" && value.trim() === "" ? undefined : value;
+
+const OptionalUrl = z.preprocess(blankToUndefined, z.string().url().optional());
+
+const OptionalNonEmptyString = z.preprocess(
+  blankToUndefined,
+  z.string().trim().min(1).optional(),
 );
 
 const OptionalGithubOrganization = z.preprocess(
@@ -69,13 +74,13 @@ const EnvSchema = z
     S3_SECRET_KEY: z.string().optional(),
     S3_BUCKET: z.string().optional(),
     AWS_REGION: z.string().optional(),
-    FACILITY_AWS_CODEBUILD_PROJECT: z.string().trim().min(1).optional(),
-    FACILITY_AWS_CODEBUILD_CACHE_BASE_LOCATION: z.string().trim().min(1).optional(),
-    VERCEL_TOKEN: z.string().trim().min(1).optional(),
-    VERCEL_OIDC_TOKEN: z.string().trim().min(1).optional(),
-    VERCEL_TEAM_ID: z.string().trim().min(1).optional(),
-    VERCEL_PROJECT_ID: z.string().trim().min(1).optional(),
-    PACKAGE_REGISTRY_TOKEN: z.string().trim().min(1).optional(),
+    FACILITY_AWS_CODEBUILD_PROJECT: OptionalNonEmptyString,
+    FACILITY_AWS_CODEBUILD_CACHE_BASE_LOCATION: OptionalNonEmptyString,
+    VERCEL_TOKEN: OptionalNonEmptyString,
+    VERCEL_OIDC_TOKEN: OptionalNonEmptyString,
+    VERCEL_TEAM_ID: OptionalNonEmptyString,
+    VERCEL_PROJECT_ID: OptionalNonEmptyString,
+    PACKAGE_REGISTRY_TOKEN: OptionalNonEmptyString,
     GITHUB_APP_ID: z.string().optional(),
     GITHUB_APP_PRIVATE_KEY: z.string().optional(),
     GITHUB_APP_WEBHOOK_SECRET: z.string().optional(),
--- a/services/api/test/config.test.ts
+++ b/services/api/test/config.test.ts
@@ -155,6 +155,22 @@ describe("API configuration", () => {
     });
   });
 
+  it("treats blank Vercel credentials from a copied .env.example as unset", () => {
+    expect(
+      readConfig({
+        ...validEnv,
+        VERCEL_TOKEN: "",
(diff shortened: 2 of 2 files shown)


mianawais78:
### Reproduction detail

Verified on Node v22.23.2 with zod 4.4.3, from `ae68401`.

**1. dotenv keeps blank keys.** Parsing the committed `.env.example`:

```
VERCEL_TOKEN         present=true value=""
VERCEL_OIDC_TOKEN    present=true value=""
VERCEL_TEAM_ID       present=true value=""
VERCEL_PROJECT_ID    present=true value=""
```

**2. The old schema rejected that.** Comparing the previous and new definitions against the three possible inputs:

```
before | absent (undefined)  -> ok
before | blank ("")          -> FAIL: Too small: expected string to have >=1 characters
before | real value          -> ok
after  | absent (undefined)  -> ok
after  | blank ("")          -> ok
after  | real value          -> ok
```

**3. It was fatal, not cosmetic.** Because `readConfig` calls `EnvSchema.parse`, the API process exited during import rather than starting with the credential unset:

```
services\api\src\config.ts:187
  const parsed = EnvSchema.parse(env);
                           ^
ZodError: [ ... "path": [ "VERCEL_TOKEN" ] ... ]
    at readConfig (services\api\src\config.ts:187:28)
    at <anonymous> (services\api\src\dev.ts:4:16)
```

All four keys were reported together, so the failure is not specific to one of them.

One reviewer question: I kept `min(1)` after normalization so a whitespace-only value such as `VERCEL_TOKEN="   "` still fails loudly, on the assumption that it indicates a broken deployment rather than an intentionally unset credential. If you would rather treat whitespace as unset too, that is a one-line change to `blankToUndefined`.

adrian-lorenzo:
Thanks for the contribution!

We’re closing this in favor of #159, which was opened earlier for the same issue and includes broader coverage for all affected variables. We appreciate the careful investigation and reproduction here.

Issue #167, opened 2026-08-16
Quickstart fails on Node 25+: corepack is no longer bundled, so corepack enable and pnpm dev cannot run


elirethDev:
### What happened

Following the quickstart in README.md and CONTRIBUTING.md on Node.js 26 (which satisfies the documented engines: node >=22):

- `corepack enable` fails: corepack is no longer bundled with Node.js 25+ (it became a separately installable package).
- `pnpm dev` then fails: pnpm is never installed, and the docs only provide the corepack path.

A fresh contributor on a current Node release cannot complete setup following the documented quickstart.

### How to reproduce

1. Install Node.js 25 or 26 (corepack not bundled).
2. git clone https://github.com/theam/facility.git && cd facility
3. corepack enable -> 'corepack' is not recognized (Windows) / command not found (Linux).
4. pnpm dev -> 'pnpm' is not recognized.

### How are you running Facility?

Local development (pnpm dev)

### Version

main (ae68401). Affected files: README.md and CONTRIBUTING.md quickstart sections.

### Evidence

- package.json declares "packageManager": "pnpm@11.20.0" and "engines": { "node": ">=22" }, so the repo metadata assumes corepack for pnpm selection while the engine range admits Node releases that no longer ship it.
- Observed on Windows 11, Node.js 26.5.1.

Suggested fix: replace `corepack enable` in the quickstart with `npm install -g pnpm@11` (works on every supported Node release), or document `npm install -g corepack` first for Node 25+ users.

PR #192, opened 2026-08-21
fix(guards): resolve markdown links against a normalised repository root

OnLocation-acumbal-contractor:
Closes #<issue number>

## The bug

`git rev-parse --show-toplevel` reports POSIX separators on every platform. On Windows it
answers `C:/repo` while `path.resolve` produces `C:\repo`, so the containment check in
`targetExists` compared

```
C:\repo\SECURITY.md   .startsWith(   "C:/repo" + "\"   )
```

which is `false`, and returned before `existsSync` was ever called. Every local link was
reported missing — on a Windows checkout `node guards/run.mjs` flags all 56 links in this
repository, `LICENSE` and `SECURITY.md` included, while `actions-pinned` passes in the same
run and the same checkout is clean on Linux.

Since `pnpm verify` runs `pnpm guards`, a Windows contributor could not get a green local
verify, and the output looks like a broken checkout rather than a platform bug.

## The fix

Normalise the root with `path.resolve` before the comparison. The containment rule is
unchanged — targets escaping the repository are still rejected.

The resolution step is extracted as an exported `resolveWithinRoot(root, source, target, path)`
with the `path` module injectable, so Windows semantics can be exercised from any host via
`node:path`'s `win32` binding. That is the only structural change; `targetExists` keeps its
behaviour and its extension/`index.md` candidates.

## Tests

`scripts/markdown-links-guard.test.mjs`, five cases, picked up by the existing
`pnpm test:dev` (`node --test scripts/*.test.mjs`):

- a git-reported Windows root (`C:/repo`) resolves in-repo targets, including a `../` hop
- a native Windows root (`C:\repo`) keeps working
- POSIX roots are unaffected
- traversal outside the repository still returns `null`, on both platforms
- the repository root itself counts as inside the repository

The traversal cases are there deliberately: the containment check is a security property, and
this change must not loosen it.

## Verification

Confirmed on a Windows host, which is where the bug is observable.

Before this change, `node guards\run.mjs` on Windows fails `markdown-links` with 56
violations — every local link in the repository, `LICENSE` and `SECURITY.md` included —
while `actions-pinned` passes in the same run. After applying this branch, same machine,
same checkout:

```
PS C:\...\facility> node guards\run.mjs
✓ actions-pinned
✓ markdown-links

2 guards ran, 0 failed.
```

```
PS C:\...\facility> node --test scripts/markdown-links-guard.test.mjs
# pass 5
# fail 0
```

On Linux:

```
node guards/run.mjs             # 2 guards ran, 0 failed
node --test scripts/*.test.mjs  # no new failures
```

## Suggested follow-up

Not in this PR, but this is the third Windows issue on the quickstart path alongside #182 and
#167. A minimal Windows CI job running `pnpm guards` and `pnpm test:dev` would have caught all
three. Happy to open that separately if it's wanted.

### DIFF
--- a/guards/markdown-links.mjs
+++ b/guards/markdown-links.mjs
@@ -1,6 +1,6 @@
 import { execFileSync } from "node:child_process";
 import { existsSync } from "node:fs";
-import { dirname, extname, resolve, sep } from "node:path";
+import nodePath, { extname, resolve } from "node:path";
 
 import { readText } from "./_kit.mjs";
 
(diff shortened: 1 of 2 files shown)


adrian-lorenzo (approved):
Thanks for the contribution!

PR #195, opened 2026-08-21
fix(kb): refuse a chain change that would strand stored entries

ophiocus:
## What changes

`PUT /v1/projects/:projectId/kb/space` now refuses a `config` whose chain would leave
an entry already stored in the space undeclared. The refusal is a `409
chain_change_strands_entries` naming the chain and every stranded entry by artifact
id; the stored config stays untouched. Doc-only saves and any config that still
declares every stored type go through exactly as before.

- `packages/harness` — `entriesStrandedByChain(entries, chain)`: the entries a chain
  leaves undeclared, by artifact id. Pure, exported, three tests of its own.
- `services/api` — the PUT handler runs it inside the update transaction against the
  entries stored for the space, before touching `patch.config`; every entry writer
  takes the space lock first (see Boundaries). Two end-to-end tests through
  `app.inject`: the refusal (write an `H` under the research default, try
  `{"chain":"product"}` → 409 with the entry named and the config unchanged; a
  charter-only save and an explicit `{"chain":"research"}` both 200), and a concurrent
  one driving both lock orderings from a second connection.
- `apps/docs` — one paragraph in *Knowledge & the Project Owner* stating the rule.

No change to `chain.ts` or to how a chain is resolved — so this sits beside #179
without overlap.

## Why

`validate()` judges every entry by the chain the space's config resolves to *now*
(`validate.ts:48,113-118`), and the PUT stored whatever config it was given
(`kb-tasks.ts:163`). A space stored with an empty config runs the research chain
(`chainFromConfig` defaults to it), so agents legitimately write H/E/F/L under it.
Rewrite that config to `{"chain":"product"}` and every one of those entries is
undeclared from then on:

- the PATCH route validates on every autosave, so the page can never be saved again —
  `unknown_artifact_type` in the editor, which is #36 as reported;
- `validateProjectKb()` validates the whole space, and the orchestrator treats its
  failure as a terminal gate (`sandbox/orchestrator.ts:321-327`), so one stranded entry
  fails the KB checkpoint of every later run in the project.

The same thing strands S/D/T/V/R in the other direction. Nothing was written wrongly in
either case; the config change re-litigated history. The fix is to refuse the change
while it would do that — which is the approach @imran-ishaq described on #179 when he
asked for this as a sibling PR, and it is the one I'd have argued for after trying the
alternative. I had a soft-allow in my pocket (let stored entries of an undeclared type
stay editable) and set it aside: it keeps pages saveable, but it treats the symptom,
and every later validation still has to know about the exception. Refusing at the
cause needs no exceptions anywhere else.

## Boundaries

- **Legacy spaces are untouched.** Nothing here changes how an existing config
  resolves; a space stored with `{}` still runs the research chain. A space that is
  *already* stranded today stays that way — recovering it (supersede or remove the
  stranded entries, or set a chain that declares them) is a separate, deliberate act,
  and the 409 tells the operator exactly which entries are involved.
- **Chain changes and entry writers are serialized.** The chain change takes the space
  row `FOR UPDATE` before deciding; every entry writer — create, supersede, intake, a
  type-changing PATCH, and the KB-amendment proposal executor — takes the same row at
  `NO KEY UPDATE` strength inside its write transaction and re-checks that the chain
  still declares its type (`lockSpaceAgainstChainChange`). Neither side can interleave
  with the other, so a stranded entry cannot slip in between the check and the config
  commit; a writer that loses the race gets `409 space_chain_changed`, not a stranded
  row. The strength matters: writers end their transactions updating the space row
  (ACTIVE upkeep), so shared locks would deadlock at that upgrade — NO KEY UPDATE
  serializes writers per space up front, which also removes the duplicate `max+1`
  numbering race on concurrent same-type creates. Covered by a concurrent integration
  test that holds the row lock from a second connection and drives both orderings.
- **The default chain for new spaces is not changed here.** The docs describe only the
  product chain, and the platform's own intake route writes `S` — yet an empty config
  resolves to research. That default is where #36 starts, and @imran-ishaq proposed
  defaulting product spaces to `chain: "product"`. It is a behaviour change for every
  new project, so I have kept it out of this PR on purpose and would open it separately
  if you want it — persisted at creation time, never at resolution time, so existing
  spaces are not re-chained retroactively (which would strand them, the exact class
  this PR closes).

## Number allocation: allocate → normalize → insert, inside the lock, at all four sites

@adrian-lorenzo caught a real one here, and the repro was the useful kind — the space
lock serialises the *insert*, but every path read `max(number)` **before** opening its
transaction, so two writers for a type carried the same number into the lock and the
second one lost on the unique constraint. Today's order was *allocate → normalize →
**lock** → insert*, with the allocation happening where nothing was serialising it.
Moving the lock boundary up in front of the allocation is what closes it, and the step
that derives from the number has to travel with it.

**The triad, applied identically at every site.** Inside the existing `db.transaction`,
right after `lockSpaceForEntryWrite` / `lockSpaceAgainstChainChange`:

1. **Allocate** — re-select `max(number)` for `(orgId, spaceId, type)` on `tx`, not `db`.
2. **Normalize** — `normalizeKbDraft` with the number that read yields.
3. **Insert** — that number, with the freshly normalised frontmatter and body.

Steps 1 and 2 now travel together as `allocateNormalizedKbEntry` in `services/api/src/harness.ts`,
so no caller can take one without the other.

### The four sites

| # | Path | File | `max` read (was) | Lock | now allocates at |
|---|---|---|---|---|---|
| 1 | `POST /v1/kb/entries/:entryId/supersede` | `services/api/src/routes/v1/kb-tasks.ts` | `559` | `609` | inside, after the lock |
| 2 | `POST /v1/projects/:projectId/kb/entries` | `services/api/src/routes/v1/kb-tasks.ts` | `688` | `745` | inside, after the lock |
| 3 | `POST /v1/projects/:projectId/kb/intake` | `services/api/src/routes/v1/kb-tasks.ts` | `1124` | `1174` | inside, after the lock |
| 4 | `executeKbAmendment` (proposal executor) | `services/api/src/executors.ts` | `2571` | `2637` | inside, after the lock |

Every site already took the lock; none of them allocated behind it. So it is the same
diff shape four times — the lock call stays put, and allocate and normalize move down to
join insert underneath it.

### Why normalize had to move too, not just the number

This was the interesting part. The number is consumed three times per site, and only the
third is the insert:

- `normalizeKbDraft` derives the artifact id from it (`H` + zero-padded number → `H003`),
  seeds `aliases` with that id, and `ensureLinks` writes `[[H003]]` into the body;
- the draft handed to `validate` carries that same id;
- the `insert` stores `number`.

So allocating late while leaving normalization early would have persisted a row whose id
contradicts its number — which `validate` rejects as `id_number_mismatch`, and which
would have been a quieter bug than the one being fixed. Hence normalize sits between
allocate and insert, under the lock.

### What survives outside the lock, and why

The pre-lock read is still there at each site, renamed to `previewNumber` with a comment
saying what it is: it dates the draft that validation inspects, and on the create route
the draft returned by `?dry=1`. It is never persisted — the stored number and the id
derived from it are always the ones allocated under the lock. Happy to move validation
inside the transaction too if you would rather have a single read; it seemed like a
bigger behavioural change than this fix needed, so I have left it as the smaller diff.

### The fifth lock holder, and why it is not a site

`kb-tasks.ts:928`, the type-changing `PATCH`, takes `lockSpaceForEntryWrite` but
allocates no number — it re-types an entry that already has one, so there is no `max`
read to move. Calling it out so its absence from the table reads as deliberate: **five
lock holders, four allocating paths** is the count to check this against.

### Regression test

`allocates distinct numbers to concurrent same-type writers`, driving your interleaving:
two `H` creates from separate connections parked behind one space lock, then released. It
asserts both succeed with consecutive numbers, and that each stored body links its own
id. Against the unfixed source it reproduces your symptom exactly — `409`, `conflict`,
*a resource with these values already exists* — which I checked by stashing only the
source change and re-running it.

## Verification

```
pnpm biome check services/api/src services/api/test   # 148 files, no fixes applied
pnpm --filter @facility/api exec tsc --noEmit         # clean
pnpm test:critical                                    # db 18/18 · api 572 passed
```

Against `main` @ `1770945` (merged in — the branch is updated by merge, never rebase),
Node 22.23.2 / pnpm 11.20.0, Postgres via `docker-compose.dev.yml`
(`facility_test` / `facility_gw`).

One thing I should flag rather than quietly leave in the numbers: after merging current
`main`, three `services/api/test/watchtower.test.ts` cases fail here — *ports outcome
math for terminal agent PRs*, *keeps merged outcomes unassessed when GitHub cannot prove
acceptance*, and *clears a stale rejection when a closed PR is reopened and merged
without evidence*. They are not from this branch: they fail identically on a detached
checkout of `1770945` with none of my changes present, so it looks like something
environmental on my side rather than anything to fix here. Shout if they are green for
you and I will dig into my setup.

Addresses #36.


services/api/src/routes/v1/kb-tasks.ts
```diff
@@ -160,7 +165,31 @@ export async function registerKbTasksRoutes(app: FastifyInstance, context: V1Rou
...
+        if (body.config !== undefined) {
+          // Re-chaining is the one config change that can strand what is
+          // already stored: validation judges every entry by the chain
+          // configured now, so an entry the new chain does not declare fails
+          // every later validation — uneditable on PATCH, and a permanent
+          // failure of the whole-space check that run checkpoints gate on.
+          // Refuse and name the entries rather than store a config that
+          // re-litigates history.
+          const next = chainFromConfig(body.config);
+          const stored = await tx
```
adrian-lorenzo, line 213: The transaction does not close the race because entry writers do not lock the space row. A writer can load the old config after this check, validate and insert an entry under that chain, while this transaction commits the new chain. That leaves exactly the stranded entry this change promises to prevent. Could we serialize chain changes with every entry-creation path and add a concurrent integration test?

services/api/src/routes/v1/kb-tasks.ts
```diff
@@ -160,7 +165,31 @@ export async function registerKbTasksRoutes(app: FastifyInstance, context: V1Rou
...
+            .select()
+            .from(kbEntries)
+            .where(and(eq(kbEntries.orgId, p.orgId), eq(kbEntries.spaceId, current.id)));
+          const stranded = entriesStrandedByChain(stored.map(toHarnessEntry), next);
+          if (stranded.length > 0) {
+            const ids = stranded.map((item) => item.artifactId).join(", ");
+            throw new ApiError(
+              409,
+              "chain_change_strands_entries",
+              `Chain "${next.id}" does not declare ${stranded.length} stored ${stranded.length === 1 ? "entry" : "entries"} (${ids}); supersede or remove them first, or keep a chain that declares their types`,
```
adrian-lorenzo: `supersede or remove them first` is not currently actionable. Superseding creates another entry of the same type and keeps the predecessor stored, so both are still reported as stranded. There also does not appear to be an entry-deletion API. Could the error suggest only a recovery path that actually works?

adrian-lorenzo (changes requested):

services/api/src/routes/v1/kb-tasks.ts
ophiocus: > The transaction does not close the race because entry writers do not lock the space row. A writer can load the old config after this check, validate and insert an entry under that chain, while this transaction commits the new chain. That leaves exactly the stranded entry this change promises to prevent. Could we serialize chain changes with every entry-creation path and add a concurrent integration test?

Right — the transaction narrowed the window without closing it; the writers never touched the row it locks. Serialized now, across two commits. (One note on how this branch is updated, since this is the first follow-up: merges and follow-up commits only, never a rebase — the diffs you reviewed stay where they were.)

- f880d8f — the chain change takes the space row `FOR UPDATE` before deciding, and the route writers (create, supersede, intake, and a type-changing PATCH) call `lockSpaceAgainstChainChange` first inside their write transaction: it locks the same row and re-checks the type they validated outside it. A writer that loses the race gets `409 space_chain_changed`, not a stranded row. The concurrent integration test drives your exact interleaving from a second connection — the writer validates under the old chain, parks at the lock while the config flips to product, is turned away, and the space still validates — plus the reverse ordering, a writer's lock parking the chain change until it commits.
- f1b1a97 — a completeness pass before re-requesting review found two more things: the KB-amendment proposal executor is a fifth writer with the same shape (validate outside, insert inside) and now takes the same lock; and the writer lock is `NO KEY UPDATE` rather than `FOR SHARE`, because two writers end their transaction by updating the space row (ACTIVE upkeep) — two SHARE holders both upgrading there is a deadlock cycle. NO KEY UPDATE serializes writers per space up front (they serialized at that tail update anyway, which also removes the duplicate `max+1` numbering race on concurrent same-type creates), never upgrades, and stays compatible with the KEY SHARE locks the entry inserts take through the space foreign key. The chain change keeps `FOR UPDATE`.
- 26aad25 — with the executor now a caller, the shared re-check no longer throws an HTTP-shaped error: it raises a domain `SpaceChainChangedError`, the routes translate it to the same 409 as before, and the executor's catch records the terse code alongside its other coded failures. No wire change; the concurrent test passes unchanged.

services/api/src/routes/v1/kb-tasks.ts
ophiocus: > `supersede or remove them first` is not currently actionable. Superseding creates another entry of the same type and keeps the predecessor stored, so both are still reported as stranded. There also does not appear to be an entry-deletion API. Could the error suggest only a recovery path that actually works?

Both points taken — superseding keeps the predecessor stored and there is no deletion API, so neither suggestion was a real way out. As of f880d8f the 409 says the one thing that is true today: the space keeps its current chain, and it can change chains only once every stored entry's type is declared by the target chain. When a genuine recovery mechanism exists (archival or deletion), the message is the right place to start offering it.

ophiocus:
Both findings addressed:

- f880d8f — serialization of chain changes against every entry writer, the concurrent two-connection test (both lock orderings), and the reworded 409.
- 47d1961 — `main` merged in (`#148`/`#192`); the branch is updated by merge, never rebase, so the reviewed history is intact.
- f1b1a97 — completeness pass: the proposal-executor writer takes the same lock, and the writer lock strength is `NO KEY UPDATE` to remove a tail-update deadlock between concurrent writers.
- 26aad25 — the shared re-check throws a host-agnostic domain error; routes translate it to the 409, the executor records the coded failure, and `harness.ts` no longer imports the HTTP error type.

`pnpm test:critical` on the branch: exit 0, no skips — db 18 · cli 98 · api 472/472 across 42 files · gateway 57. PR description updated to match. Re-requesting review.

adrian-lorenzo (changes requested):
Thanks for the contribution!

The new space lock does not remove the concurrent `max + 1` numbering race because each writer calculates the maximum before entering the transaction and taking the lock. I reproduced this by parking two `H` creates behind the same space lock; after releasing it, one succeeded and the other returned `409` because both had selected the same number.

Please allocate the number after acquiring the lock in every affected path—create, supersede, intake, and the KB amendment executor—and add a regression test with concurrent same-type writers.

With that fixed, we can approve and merge it!

ophiocus:
Great catch, and thank you for the repro — parking two `H` creates behind the lock is exactly the test I should have written the first time. You are right on both counts: the lock serialised the insert, not the allocation, and my earlier note claiming it also settled the numbering was wrong. Every path read `max(number)` before opening its transaction, so both writers walked in carrying the same number.

Fixed in `027720d`, as the same move at each of the four allocating paths — allocate, normalize, insert, all inside the lock:

| # | Path | File |
|---|---|---|
| 1 | `POST /v1/kb/entries/:entryId/supersede` | `services/api/src/routes/v1/kb-tasks.ts` |
| 2 | `POST /v1/projects/:projectId/kb/entries` | `services/api/src/routes/v1/kb-tasks.ts` |
| 3 | `POST /v1/projects/:projectId/kb/intake` | `services/api/src/routes/v1/kb-tasks.ts` |
| 4 | `executeKbAmendment` | `services/api/src/executors.ts` |

Chasing it turned up something I had not expected, which is why the diff is a little larger than "move one query": the number is consumed three times per site and only the last is the insert. `normalizeKbDraft` derives the artifact id from it (`H003`), seeds `aliases` with that id, and `ensureLinks` writes `[[H003]]` into the body — so allocating late while normalizing early would have stored a row whose id contradicts its number, which `validate` then rejects as `id_number_mismatch`. Allocation and normalization now travel together as `allocateNormalizedKbEntry` in `harness.ts`, so no caller can take one without the other.

Two notes on what I did *not* do, in case you would rather I did:

- The pre-lock read survives at each site as `previewNumber`, commented as such — it dates the draft validation inspects, and on the create route the `?dry=1` response. It is never persisted. Moving validation inside the transaction as well would give you a single read; that felt like a bigger behavioural change than this fix needed, but say the word and I will.
- `kb-tasks.ts:928`, the type-changing `PATCH`, is untouched: it holds the same lock but allocates nothing, since it re-types an entry that already has a number. Five lock holders, four allocating paths — flagging it so its absence reads as deliberate rather than missed.

The regression test is `allocates distinct numbers to concurrent same-type writers`, driving your interleaving from two connections and asserting both succeed with consecutive numbers and each body linking its own id. I checked it actually bites by stashing only the source change and re-running: it reproduces your symptom exactly — `409`, `conflict`, *a resource with these values already exists*.

`pnpm test:critical` is green here apart from three `watchtower.test.ts` cases that also fail on a detached checkout of `1770945` with none of my changes present, so I have written them up in the PR body as environmental on my side rather than something to fix here — do let me know if they are green for you.

Branch is still merge-only, never rebased, so the diffs you have already reviewed have not moved. Thanks again for the careful read on this one — it is a better change for it.

adrian-lorenzo (approved):
Thanks for the careful follow-up—the concurrency fix is solid!

adrian-lorenzo (approved):

PR #196, opened 2026-08-22
fix(api): include canonical query parameters in POST idempotency fingerprint

### LINKED ISSUE #187: POST idempotency ignores query parameters that change request behavior
## The problem

Facility currently scopes authenticated POST idempotency by principal, method, query-stripped path, idempotency key, and request body.

This matches the current API documentation, which says that the same principal/method/path/key and body replays the stored response. However, some POST query parameters change the operation’s behavior or target. Excluding those parameters can cause one operation to be silently replayed for a semantically different operation.

I reproduced this using the KB entry endpoint:

- `POST /v1/projects/:projectId/kb/entries?dry=1` validates an entry without persisting it.
- `POST /v1/projects/:projectId/kb/entries` validates and persists the entry.

I sent the dry-run request first, followed by the persistent request with the same fresh `Idempotency-Key` and JSON body.

Observed result:

```text
Dry status:        200
Dry idempotency:   created
Live status:       200
Live idempotency:  replayed
Persisted entries: 0
```

The persistent operation was treated as a replay of the validation-only operation. It returned a successful response, but its intended write never occurred.

Reusing an idempotency key for semantically different requests is a client error. The contract question is whether Facility should detect that error when the semantic difference is carried in the query, as it already does when the request body changes.

## How to reproduce

1. Start Facility locally.
2. Create a disposable project with a KB space.
3. Create a project-scoped API key with `kb:write`.
4. Prepare a valid `H` entry:

   ```json
   {
     "type": "H",
     "slug": "idempotency-query-reproduction",
     "frontmatter": {},
     "bodyMd": "Idempotency query reproduction",
     "links": []
   }
   ```

5. Generate a fresh idempotency key.
6. Send the validation-only request:

   ```http
   POST /v1/projects/<project-id>/kb/entries?dry=1
   Authorization: Bearer <redacted>
   Idempotency-Key: <fresh-key>
   Content-Type: application/json
   ```

7. Send the persistent request with the same body and key:

   ```http
   POST /v1/projects/<project-id>/kb/entries
   Authorization: Bearer <redacted>
   Idempotency-Key: <same-key>
   Content-Type: application/json
   ```

8. List the project’s KB entries and search for the submitted slug.

The first response has `Idempotency-Status: created`. The second has `Idempotency-Status: replayed`, and the entry is absent from the KB.

## Implementation evidence

The runtime result is consistent with the implementation:

- [`services/api/src/idempotency.ts:32-37`](https://github.com/theam/facility/blob/6060ab51267c61aaf787f1cfa226cb4a9fabb9b2/services/api/src/idempotency.ts#L32-L37) removes the query string from the path, hashes only `request.body`, and builds the record identity without the query.
- [`services/api/src/routes/v1/kb-tasks.ts:580-618`](https://github.com/theam/facility/blob/6060ab51267c61aaf787f1cfa226cb4a9fabb9b2/services/api/src/routes/v1/kb-tasks.ts#L580-L618) defines `dry` and uses it to switch between validation-only and persistence behavior.
- [`apps/docs/docs/reference/api.md:38-42`](https://github.com/theam/facility/blob/6060ab51267c61aaf787f1cfa226cb4a9fabb9b2/apps/docs/docs/reference/api.md#L38-L42) documents replay using principal, method, path, key, and body, but does not state how behavior-changing query parameters should be handled.

Source inspection suggests the same class may affect POST routes where a query such as `repoId` selects the operation’s target. I have only runtime-reproduced the KB dry/live case.

## Proposed behavior

Define behavior-changing query parameters as part of POST idempotency equivalence.

A possible contract would be:

- The same key, path, semantically equivalent query, and body replays the stored response.
- Reusing the same key with a different query returns `409 idempotency_key_reused`, consistent with changed-body handling.
- Query parameters are canonicalized before comparison so ordering differences do not create false conflicts.
- The API documentation explicitly states how query parameters participate in idempotency equivalence.

The existing principal/method/path/key identity could remain unchanged while a normalized representation of the query participates in the request fingerprint. The exact representation is an implementation decision; the important behavior is that semantically different operations do not silently replay one another.

### Require clients to use different keys

Clients can use separate keys for validation and persistence. This works under the current contract, but relies on every caller recognizing that a query change creates a different logical operation. Facility already detects this misuse when the body changes but not when the query changes.

### Include the raw query string

This would distinguish the operations, but parameter ordering and equivalent encodings could create false conflicts. Canonicalizing the parsed query would be safer.

### Move `dry` into the body or use a separate endpoint

Either approach would make the current body/path fingerprint sufficient, but it would be a larger API contract change.

### Document the existing behavior only

If excluding queries is deliberate, the documentation could state that explicitly and client helpers could ensure validation and persistence never share a key. This is smaller, but the silent replay risk remains.

## Acceptance criteria

- Identical POST requests with the same meaningful query, body, and key still replay.
- Reusing the same key with a behavior-changing query difference does not silently replay.
- Equivalent query parameter orderings are treated consistently.
- The API documentation describes the chosen query-equivalence behavior.
- Regression coverage includes the KB `dry=1` followed by persistent-create case.

## Environment

- Local development with `pnpm dev`
- Tested commit: `6060ab51267c61aaf787f1cfa226cb4a9fabb9b2`

I would be happy to contribute a focused implementation and regression test once the intended query-equivalence contract is confirmed.

Ganador1:
Closes #187

## Why

As reported in #187, Facility scopes authenticated POST idempotency by principal, method, query-stripped path, key hash, and request body hash. 

However, some POST endpoints use query parameters to alter operation semantics (for example, `POST /v1/projects/:id/kb/entries?dry=1` performs validation-only without persistence, while `POST /v1/projects/:id/kb/entries` performs live persistence).

Because `requestHash` only hashed `request.body`, a dry-run validation request followed by a live persistent request reusing the same `Idempotency-Key` was falsely treated as a replay. The API returned a successful 200 response with `idempotency-status: replayed`, but the intended write never occurred.

## What it does

1. **Includes canonical query parameters in the request fingerprint**: Computes `requestHash` as `sha256(`${stableQuery}|${stableBody}`)`, where `stableQuery` is a deterministic, key-sorted JSON serialization of `request.query`.
2. **Detects behavior-changing query parameter differences**: Reusing an `Idempotency-Key` with different query parameters is detected as a mismatch and rejected with `409 idempotency_key_reused`, matching the existing body-conflict behavior.
3. **Preserves parameter-order tolerance**: Permuted query parameters (e.g. `?a=1&b=2` vs `?b=2&a=1`) produce identical canonical representations and replay correctly without false conflicts.
4. **Fails closed for unversioned legacy fingerprints**: Replays against legacy records (`sha256(stableJson(body))`) fail closed with `409 idempotency_key_reused` because legacy records are inherently ambiguous and cannot prove whether the original request contained query parameters (such as `?dry=1`).
5. **Updates API documentation**: Clarifies in `apps/docs/docs/reference/api.md` that canonical query parameters participate in idempotency equivalence.
6. **Adds comprehensive test coverage**:
   - `services/api/test/idempotency.unit.test.ts`: Unit test suite validating initial claims, permuted query replays, key reuse rejections with different query parameters, and fail-closed rejections on legacy dry-run vs live writes.
   - `services/api/test/api.test.ts`: Integration tests covering query permutation replays, query difference rejections, and the exact KB dry vs persistent create reproduction from #187.

## Verification

Commands run:

```bash
pnpm lint                                                               # Biome clean across 423 files
pnpm typecheck                                                          # 16 packages passed, 0 errors
node guards/run.mjs                                                     # 2 guards ran, 0 failed
pnpm --filter @facility/api exec vitest run test/idempotency.unit.test.ts # 5/5 passed (4ms)
pnpm test:uncached                                                      # 10 test suites passed
```


jungle-lethanh:
Thanks for linking #187 and covering the KB dry/live reproduction. I had deliberately held off on implementation while waiting for maintainer confirmation of the intended query-equivalence contract.
One rollout case may be worth checking: existing idempotency records were fingerprinted as `sha256(stableJson(body))`, while this change fingerprints `stableQuery|stableBody`. A retry using a record created before deployment could therefore return `409 idempotency_key_reused` for up to 24 hours even when the request is otherwise unchanged.
Is that compatibility behavior intentional? If so, it may be worth documenting. Otherwise, a regression test or transitional fingerprint strategy may be needed.
I’m also happy to validate the final behavior against the original runtime reproduction.

services/api/src/idempotency.ts
```diff
@@ -31,7 +31,9 @@ export async function beginIdempotentRequest(
   if (!principal) throw new ApiError(401, "unauthorized", "Authentication required");
   const path = request.url.split("?")[0] ?? request.url;
   const keyHash = sha256(key);
-  const requestHash = sha256(stableJson(request.body));
+  const stableQuery = stableJson(request.query ?? {});
+  const stableBody = stableJson(request.body);
+  const requestHash = sha256(`${stableQuery}|${stableBody}`);
```
adrian-lorenzo, line 36: Existing idempotency records contain the body-only hash and remain valid for 24 hours. After deploying this change, an identical retry against one of those records will compute a different hash and return `409 idempotency_key_reused`. This can also happen between old and new instances during a rolling deployment. Could we preserve compatibility with existing records and add a regression test for this case?

adrian-lorenzo (changes requested):

services/api/src/idempotency.ts
Ganador1: Good catch, @adrian-lorenzo. I've updated the implementation in commit 292df7c to support backward compatibility:

- When verifying `existing.requestHash`, we also check against `legacyRequestHash = sha256(stableJson(request.body))`.
- If an existing record was created prior to deployment (or by an older instance during a rolling deployment) with the body-only hash, an identical retry matches `legacyRequestHash` and replays seamlessly.
- Replays with different payloads continue to be rejected with `409 idempotency_key_reused`.
- Added unit regression coverage in `idempotency.unit.test.ts` for legacy record replays.

adrian-lorenzo (changes requested):
Thanks for the contribution!

The legacy fallback still allows the original bug during a rolling deployment. A legacy record stores only the body hash, so it contains no information about the query that created it. `existing.requestHash === legacyRequestHash` therefore accepts any query with the same body. I reproduced this by retrying a legacy record with the same body plus `?dry=1`; it was replayed instead of returning `409`.

Please don’t treat a body-only legacy hash as proof that the query matches. Legacy records are inherently ambiguous, so the transition should fail safely or store explicit fingerprint version/query information. Please also add a regression test for a legacy record receiving the same body with a changed query.

With that fixed, we can approve and merge it!

Ganador1:
Thanks for the thorough review and catch, @adrian-lorenzo!

I've updated the implementation in commit 9e1c536 to fail safely:

1. **Strict Queryless Legacy Replays**: A body-only legacy hash match is now strictly restricted to incoming requests without query parameters (`stableQuery === "{}"`).
2. **Fail-Safe on Changed Queries**: If an incoming request introduces or modifies query parameters (such as `?dry=1`), it does not match the legacy body-only hash and fails safely with `409 idempotency_key_reused`.
3. **Regression Test Added**: Added a test case in `services/api/test/idempotency.unit.test.ts` verifying that attempting to retry an existing legacy record with the same body but with `?dry=1` is rejected with `409 idempotency_key_reused`.

adrian-lorenzo (changes requested):
Thanks for following up on the review!

The legacy ambiguity remains in the opposite direction. A legacy record may have been created by a request containing `?dry=1`; retrying that key and body without a query makes `stableQuery === "{}"`, so `isLegacyQuerylessMatch` accepts the record and replays the dry response instead of performing the live write. That is the original failure mode.

Please fail closed for every unversioned legacy fingerprint or add explicit fingerprint-version/query information. Add a regression that seeds a legacy dry-run response, sends the queryless live request with the same key and body, and verifies that it is rejected rather than replayed.

Once that case is covered, we can approve it.

Ganador1:
Thanks for clarifying the reverse ambiguity case, @adrian-lorenzo! That makes total sense: because legacy records stored only `sha256(body)`, a record originally created with `?dry=1` would have replayed its dry response for a subsequent live queryless write, masking the write.

I've updated the implementation in commit fdf259b to **fail closed** for all unversioned legacy fingerprints:

1. **Fail-Closed on Unversioned Legacy Fingerprints**: Any record that does not match the canonical query+body `requestHash` is rejected with `409 idempotency_key_reused`. We no longer attempt fallback matching against ambiguous body-only legacy hashes.
2. **Regression Test for Legacy Dry-Run vs Live Write**: Added a test in `services/api/test/idempotency.unit.test.ts` that seeds a legacy dry-run response (`{ ok: true, validationOnly: true }` fingerprinted with the body-only hash), sends a live queryless request with the same key and body, and asserts that it is rejected with `409 idempotency_key_reused` rather than replayed.
3. **Full Test Suite & Linter Passing**: All unit and monorepo suites are passing cleanly.

adrian-lorenzo (approved):
Thanks for addressing the feedback!

Lob26:
The fingerprint change reads correct to me: sorting keys before hashing keeps permuted parameters equivalent, and treating a differing query as `409 idempotency_key_reused` matches what the body already does rather than inventing a second behaviour for the same class of conflict.

Two things I checked against `main` that I think the description should carry, because this has been open a while and the tree moved underneath it.

**The motivating endpoint is gone.** `kb/entries` no longer exists; the knowledge base went with the 0.12 rebuild. I also could not find a single route that combines `idempotent: true` with a `querystring` schema on `main` today, so as far as I can tell no current route can reach the bug. That is not an argument against the change, it is an argument for reframing it: preventive hardening on a fingerprint that is currently one query parameter away from being wrong is an easy yes, whereas a bug report pointing at a deleted endpoint invites a close.

**The fail-closed window is worth stating.** Rejecting legacy fingerprints is the right call, and `IDEMPOTENCY_TTL_MS` is 24 h, so it self-heals. The window that actually matters is smaller and different: API replicas are deployed rolling, so during the overlap a record claimed by old code and replayed against new code answers 409 instead of replaying. Naming that in the PR body, with the observation that it is bounded by the deploy overlap and not by the 24 h TTL, turns the scariest-sounding part of the change into a bounded risk a reviewer can accept.

Happy to help with either the reframing or a regression against a route that exists today, if that is useful. It is your change and I do not want to take it over.

PR #197, opened 2026-08-22
feat(receipts): name non-gating evidence a run failed to preserve

Arifuzzamanjoy:
When a transcript upload fails the run still reports succeeded. It reaches the
receipt as an anonymous +1 on activity.errors, so the record can say a run hit
an error but never that the transcript is what it lost.

The runner emits a named evidence event: type "evidence", name transcript,
reason transcript_upload_failed. Not a platform check — every
self_reported: false failed check in runner/src/index.ts feeds
engineAndChecksSucceeded, and one that did not would let a run succeed while
its receipt carries a failed platform check. The receipt collects type
"evidence" alongside its checks, with truncation disclosed the same way
checks_truncated is.

artifact_error stays and keeps feeding the error count, so the existing tally is
unchanged and the runner and API can deploy in either order.

The emit is best-effort. emit() throws on a non-429 response and that reaches
main()'s outer catch, which posts a failed result. An unguarded emit would fail
a successful run over a storage blip, which is the gating this event exists to
avoid.

What it does not do is gate the run. Gating would skip the delivery phase over a
transient storage failure and throw away correct agent work to protect its own
paper trail. That trade belongs to you, so I stopped at making the loss visible.

Worth naming for two reasons: the api and security docs describe transcripts as
durably stored, so a silent loss contradicts that; and in read-only modes the
transcript is the only evidence a run produces, since nothing gets committed.

Touches runner/src/index.ts, packages/core/src/receipts.ts, and
services/api/src/sandbox/orchestrator.ts, with tests in runner,
packages/core, and services/api/test/sandbox.test.ts.

### DIFF
--- a/packages/core/src/receipts.ts
+++ b/packages/core/src/receipts.ts
@@ -8,6 +8,15 @@ const ReceiptCheckSchema = z.object({
   exit_code: z.number().int().optional(),
 });
 
+// Non-gating evidence: named artifacts a run was expected to preserve. Distinct
+// from `checks`, which the run's outcome is gated on. Optional so receipts
+// sealed before this field existed still parse and digest unchanged.
+const ReceiptEvidenceSchema = z.object({
+  name: z.string(),
+  status: z.enum(["passed", "failed", "skipped", "unknown"]),
+  reason: z.string().optional(),
+});
+
 export const FacilityReceiptSchema = z.object({
   schema: z.literal("facility.run.v1"),
   run_id: z.string().optional(),
@@ -71,6 +80,8 @@ export const FacilityReceiptSchema = z.object({
     .optional(),
   checks: z.array(ReceiptCheckSchema).optional(),
   checks_truncated: z.boolean().optional(),
+  evidence: z.array(ReceiptEvidenceSchema).optional(),
+  evidence_truncated: z.boolean().optional(),
   integrity: z
     .object({
       algorithm: z.literal("sha256"),
--- a/packages/core/test/core.test.ts
+++ b/packages/core/test/core.test.ts
@@ -468,4 +468,53 @@ describe("receipts", () => {
       }),
     ).toBe(false);
   });
+
+  it("seals non-gating evidence without disturbing receipts that carry none", () => {
+    const base = parseLegacyAgentReceipt({
+      schema: "example.agent_sdlc.run.v1",
+      provider: "claude_code",
+      mode: "builder",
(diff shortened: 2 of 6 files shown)


adrian-lorenzo (changes requested):
Thanks for the contribution! `self_reported: false` makes this a platform check, but it does not feed `engineAndChecksSucceeded`, so the run can succeed while its receipt contains a failed platform check. That contradicts the runner’s invariant that every platform check must pass and gives consumers conflicting results. Please either make transcript upload failure gate the result, or represent it as a separate non-gating evidence event. Please also test the final result/receipt behavior rather than only the helper’s output.

Arifuzzamanjoy:
went with the evidence event. type: "evidence", no self_reported, so it cannot
land in the check list and contradict the result.

added runner/test/transcript-upload.integration.test.ts. it asserts the state
uploadTranscript returns, and that filtering emitted events by type = "check",
the way the receipt query does, never picks this one up.

one thing that follows from that. receipts only read type = "check", so the
receipt itself does not change. the loss is named in run_events but not shown on
the receipt. let me know if you want it shown there.

Arifuzzamanjoy:
Rebased onto current upstream. ready for re-review.

Arifuzzamanjoy:
Receipt now collects the evidence event, so it names the loss instead of sitting
in run_events doing nothing. tested through POST /result against the stored
receipt. that reaches packages/core and services/api, flagging it rather than
slipping it past you.

Retitled. say the word if you'd rather it read feat(receipts).

adrian-lorenzo (changes requested):
Thanks for the contribution!

Transcript failure is currently recorded through two separate event requests: the anonymous `artifact_error`, followed by the named evidence. If the first succeeds and the second hits a transient failure, the receipt keeps only the anonymous error—the behavior this PR is intended to fix. I reproduced this with responses of `200` followed by `503`.

Please send both events in one batch and add a regression test for this partial-failure case. The test should also exercise the real capture path rather than recreating its wiring manually, so removing the call from `main()` cannot leave the tests green.

Arifuzzamanjoy:
Both events go out in one emit(), and appendRunEvents inserts a batch in a single
transaction, so the anonymous half can't land alone. transcript-upload.integration.test.ts
covers the 200-then-503 case and asserts what the platform stored; a call-site
assertion fails if main() stops calling the phase, since main() itself isn't
exported.

PR #206, opened 2026-08-28
refactor(config): share the blank-to-undefined env preprocess

christian-macedo:
Focused refactor of `services/api/src/config.ts`. Builds on **#159**, which fixed the empty-optional-env crash and added the regression coverage — this PR changes no behavior and adds no tests.

## What this does

After #159, the empty-string→undefined preprocess lambda was copied inline into four optional env declarations:

- `OptionalUrl`
- `OptionalGithubOrganization`
- `OptionalNonEmpty` (the seven `VERCEL_*` / `FACILITY_AWS_CODEBUILD_*` / `PACKAGE_REGISTRY_TOKEN` fields)
- `FACILITY_PREVIEW_SURFACE_TOKEN`

This extracts that idiom into a single `blankToUndefined` helper and reuses it across all four, so the behavior has one definition instead of four copies.

## What this deliberately does not do

- **No behavior change** — the parse result for every field is identical.
- **No new tests** — the blank-optional-env regression coverage already lives in #159; the existing `config.test.ts` (34 cases, including those) continues to pass unchanged.

## Scope

- `services/api/src/config.ts` — +10 / −13, single file.

## Verification

- `vitest run test/config.test.ts` — 34 passed (includes #159's blank/whitespace regression cases)
- `biome check services/api/src/config.ts` — clean

`refactor:` — no release impact on its own, matching the no-behavior-change scope.


adrian-lorenzo:
Thanks for the contribution! The root-cause analysis is correct, and centralizing the blank-value normalization in `blankToUndefined` is a useful cleanup.

#159 already implements and has approval for the same functional fix. Please reframe this PR as a focused configuration refactor that keeps only the shared helper and its reuse across the existing optional schemas, without duplicating #159’s behavior and regression coverage.

That would preserve the distinct maintainability improvement from this contribution.

adrian-lorenzo (approved):
Thanks for refining the contribution!

christian-macedo:
> Thanks for refining the contribution!

@adrian-lorenzo can I please get the approvals on the workflows so I can merge the changes.

PR #214, opened 2026-08-29
feat(cli): detect project roots across ecosystems, not just Node

### LINKED ISSUE #199: Stack detection covers only the Node ecosystem — every other stack starts on an empty job site
## Problem

`facility init` promises to detect the repository's package manager and checks, but
`packages/cli/src/detect.mjs` only knows the Node ecosystem:

- Package manager: `pnpm-lock.yaml` → `yarn.lock` → `package-lock.json` → `package.json` → **`"none"`** (`detect.mjs:26-35`). There is no probe for `composer.json`, `pyproject.toml`, `go.mod`, `Cargo.toml`, `pom.xml`/`build.gradle`, or `Gemfile`.
- Checks: derived exclusively from `package.json` scripts (`typecheck`/`lint`/`test`/`build`), so a non-Node repo gets an **empty check list**.
- `migrationDirs` knows only `migrations`, `supabase/migrations`, `db/migrations`, `prisma/migrations` — no `database/migrations` (Laravel), no Alembic/Django layouts.
- The generated workflow for a non-Node repo carries a stub: `# facility: no Node toolchain detected. Add your language setup steps` (`init.mjs:54`).

## Why it matters

The method doc names the dominant failure mode Facility exists to prevent: *"the
environment can't run anything, so verification is impossible, so the agent hedges …
The fix is not a better prompt. It is a better job site."* Today, every PHP, Python,
Go, Java, Ruby, and Rust repository begins its Facility life on exactly that empty
job site: no provision proposal, no checks, a comment where the toolchain should be.
The delivery contract ("run the configured checks; a partial deliverable is a
failure") has nothing to bind against, and the first `/builder` run produces the
hedging behavior the method predicts.

The current behavior also fails silently from the adopter's point of view: `init`
completes, the questions get asked, and the gap only surfaces at the first paid run
— the same late-surfacing shape as #28.

## Proposal

A deterministic ecosystem adapter table in `detect.mjs`, the same pattern as
`detectDeploymentProvider`:

| Signal | Provision proposal | Check proposals | Migration dirs |
|---|---|---|---|
| `composer.json` | `composer install` | `vendor/bin/phpunit` · `vendor/bin/pest` · `composer test` script if declared; `vendor/bin/phpstan`/`psalm` if configured | `database/migrations` |
| `pyproject.toml` / `requirements*.txt` | `pip install -e .[dev]` / `uv sync` | `pytest`; `ruff check` if configured | `alembic/`, `*/migrations/` |
| `go.mod` | `go mod download` | `go vet ./...` · `go test ./...` | `migrations/` |
| `Cargo.toml` | — | `cargo clippy` · `cargo test` | `migrations/` |
| `pom.xml` / `build.gradle` | — | `mvn -q verify` / `gradle test` | Flyway/Liquibase dirs |
| `Gemfile` | `bundle install` | `bundle exec rspec` / `rake test` | `db/migrate` |

Scope boundaries, to keep it reviewable:

- **Proposals, not assumptions** — everything stays a default the `init` prompts
  show for confirmation, exactly as the Node defaults do today. Repos that declare
  their commands keep them; detection only fills silence.
- One coherent change to `detect.mjs` + its fixture-repo tests (the CLI test
  pattern already runs the real installer against temporary repositories).
- No new dependencies (CLI stays zero-runtime-dependency).
- Multi-ecosystem repos (e.g. PHP backend + Node frontend) record all matches and
  propose the union of checks, mirroring how deployment providers accumulate.

I'm happy to implement this behind the issue-first agreement in CONTRIBUTING.md,
starting with the `composer.json` adapter (PHP is where I can validate against real
repositories with real provisioning — databases, migrations, coverage drivers —
rather than toy fixtures) and adding the others in the same PR if the shape is
agreed, or in follow-ups if you'd rather keep the first diff minimal.

ophiocus:
Closes #199 — and big thanks to @AndreyGritsa, whose nine-Poetry-roots-plus-Next.js layout from the issue thread is now a test case in this PR. That fixture reshaped the design: the fix isn't an adapter table alone, it's **multi-root, multi-ecosystem** detection, because a repository with no root `package.json` and independent Python packages beside a frontend was previously invisible to `init`.

## What changed

`packages/cli/src/detect.mjs`:

- **Ecosystem adapter table** — Node, Python (Poetry/uv/PDM/Hatch/pip), PHP (Composer), Go, Ruby, Rust, Java (Maven/Gradle). Same shape as `detectDeploymentProvider`: adding an ecosystem is one entry, not a branch. Each adapter only proposes tools the project actually configures (ruff/mypy/pytest from `pyproject.toml`, phpunit/pest from the files present, and so on).
- **Multi-root walk** (depth ≤ 3) — every project root is discovered and kept. Checks are unioned and root-scoped — `(cd scraper && poetry run pytest)` — so a change anywhere in the repository has a gate. Provisioning is per root (`poetry install` in the touched package), not one repo-wide install, which a layout like Andrey's can't express.
- **Workspace absorption** — pnpm/npm/Cargo workspace members (and packages nested below them, e.g. `runner/agent-clis`) are absorbed into their workspace root, whose scripts already fan out to them. This is what keeps existing Node repositories detecting **identically to before**.
- **Fixture/sample directories are never roots** — `test/`, `fixtures/`, `samples/`, `examples/` hold manifests on purpose; proposing their checks as repository defaults would be noise (found by running the new detection against claude-dev-kit, where `samples/php` volunteered its phpunit as a repo default).
- **`test:run` fallback** when `test` is absent — the watch-free vitest convention, and literally the script name in the frontend from the issue thread.

`detect()` also returns two new advisory fields, `ecosystems` and `roots` (path/ecosystem/manager), which `init` can surface in its summary; existing consumers are untouched.

`apps/docs/docs/faq.md`: the "Does this work for non-Node projects?" answer now describes what detection actually does instead of promising only a marked slot.

## How it's verified

- New `packages/cli/test/detect.test.mjs`, 5 cases: single-root npm unchanged; pnpm workspace absorption incl. the nested-member case; the polyglot no-root-manifest monorepo from #199; PHP + Go roots with migration discovery; fixture/sample dirs never become roots.
- Full CLI suite: **103 tests, 0 failures** (98 before this PR — the five are new).
- **Backward-compatibility proof, not just tests**: ran `main`'s `detect()` and this branch's `detect()` against facility's own checkout — every shared output field is identical, byte for byte. CDK's real tree confirms the fixture-skip.

@AndreyGritsa — your offer to validate against the real monorepo stands very welcome: this should now discover all nine Poetry roots plus `frontend/`, propose `poetry install` per touched package, and union `poetry run pytest` with `npm run test:run`. I'd love to hear where it guesses wrong.

### DIFF
--- a/apps/docs/docs/faq.md
+++ b/apps/docs/docs/faq.md
@@ -59,9 +59,13 @@ the code in your repo.
 ## Does this work for non-Node projects?
 
 Yes. The CLI and the vendored guards need Node on the runner (present on all
-GitHub-hosted runners) — your project doesn't. `init` detects pnpm/yarn/npm
-and otherwise leaves a marked slot for your toolchain steps. The provision
-command is yours: `make db`, `docker compose up -d`, `mix ecto.setup`.
+GitHub-hosted runners) — your project doesn't. `init` detects project roots
+across ecosystems — Node (pnpm/yarn/npm), Python (Poetry/uv/PDM/Hatch/pip),
+PHP (Composer), Go, Ruby, Rust, and Java — including several side by side in
+one repository, and proposes per-root provision and check commands. Anything
+it cannot see still gets a marked slot for your toolchain steps, and the
+provision command stays yours: `make db`, `docker compose up -d`,
+`mix ecto.setup`.
 
 ## Can I use my existing AGENTS.md / CLAUDE.md / .claude setup?
 
--- a/packages/cli/src/detect.mjs
+++ b/packages/cli/src/detect.mjs
@@ -93,12 +93,20 @@ const ADAPTERS = [
       }
       // The toolchain step already installs Node dependencies, so provisioning
       // here means only the project's own extra setup.
+      const install = existsSync(join(dir, "pnpm-lock.yaml"))
+        ? "pnpm install --frozen-lockfile"
+        : existsSync(join(dir, "yarn.lock"))
+          ? "yarn install --immutable"
+          : existsSync(join(dir, "package-lock.json"))
+            ? "npm ci"
+            : "npm install";
       const pnpmWorkspace = readText(join(dir, "pnpm-workspace.yaml"));
       const declared = Array.isArray(pkg.workspaces) ? pkg.workspaces : (pkg.workspaces?.packages ?? []);
       const workspaceGlobs = pnpmWorkspace ? pnpmWorkspaceGlobs(pnpmWorkspace) : declared;
 
       return {
         manager,
+        install,
         provision: scripts.setup ? `${runner} setup` : "",
         checks,
         workspaceGlobs,
@@ -341,8 +349,15 @@ function detectRoots(dir) {
 
 // A command for a root below the repository root has to run there. Roots at
 // "." are emitted verbatim, so single-root repositories are unchanged.
+// Discovered directory names come from repository contents and may legally
+// contain shell metacharacters, so the path is always single-quoted (with
+// embedded quotes escaped) before it reaches a shell command.
+function shellQuote(value) {
+  return "'" + value.replace(/'/g, "'\\''") + "'";
+}
+
 function scopeCommand(rootPath, command) {
-  return rootPath === "." ? command : `(cd ${rootPath} && ${command})`;
+  return rootPath === "." ? command : `(cd ${shellQuote(rootPath)} && ${command})`;
 }
 
 export function detect(dir) {
@@ -361,8 +376,15 @@ export function detect(dir) {
       const scoped = scopeCommand(root.path, check);
       if (!checks.includes(scoped)) checks.push(scoped);
     }
-    if (root.provision) {
-      const scoped = scopeCommand(root.path, root.provision);
+    const provisionParts = [];
+    if (root.ecosystem === "node" && root.path !== "." && root.install) {
+      // A nested Node root is outside the CI toolchain steps' install, so its
+      // checks would otherwise run against an empty node_modules.
+      provisionParts.push(root.install);
+    }
+    if (root.provision) provisionParts.push(root.provision);
+    for (const part of provisionParts) {
+      const scoped = scopeCommand(root.path, part);
       if (!provisions.includes(scoped)) provisions.push(scoped);
     }
   }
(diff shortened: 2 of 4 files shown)


adrian-lorenzo (changes requested):
Thanks for the substantial contribution!

`scopeCommand()` places discovered directory names directly into shell commands. These names come from repository contents and may legally contain shell metacharacters. I reproduced command substitution with a hostile directory name, so please shell-quote these paths and add a regression test covering the generated commands.

There is also a functional gap for nested Node roots: a repository containing only `frontend/package.json` produces `packageManager: "none"` and no provision command, while its checks still run from `frontend`. Please ensure dependencies are installed for nested Node roots and cover this layout end to end.

This is a strong direction; once these cases are fixed, it should be ready for another pass.

ophiocus:
Both addressed in 9452354 — thank you for reproducing the substitution rather than just flagging it; that made the fix precise.

- **Quoting:** `scopeCommand()` now single-quotes every discovered path with embedded quotes escaped (`'don'\''t'`), and the regression test pins both the metacharacter case — a directory literally named `evil$(echo pwned)` stays inert — and the embedded-quote case.
- **Nested Node roots:** a repo whose only Node root is `frontend/` now gets a lockfile-aware install in the provision chain (`(cd 'frontend' && npm ci)`; pnpm/yarn variants likewise, `npm install` when no lock exists), covered by a dedicated test. Root-level behavior is untouched — the CI toolchain steps keep owning the root install, so existing repos detect byte-identically.

Full CLI suite: 105 tests, 0 failures on Node 24.

ophiocus:
One more commit to honor the full sentence of your review — "cover this layout end to end" deserved more than unit tests. cbb13de adds the true e2e: `init --yes` runs against a repository whose only Node root is `frontend/`, and the test asserts the install is rendered into the shipped artifacts themselves — `.facility.json` carries `(cd 'frontend' && npm ci)` in its provision and the crew workflow carries it ahead of the scoped checks. Suite now 106 tests, 0 failures. That should close both halves.

ophiocus:
Closing this one. #289 rewrote `detect.mjs` down to a Node-only read and moved repository setup into `.facility.yml`, so root detection across ecosystems is no longer the lever it was when this was filed; re-adapting it onto the new file would be a third rewrite of a feature 0.12 made optional. Thanks for the review, @adrian-lorenzo — the point that discovered directory names come from repository contents and must be shell-quoted before they reach a command (you reproduced the substitution with a hostile name) is the one to carry into anything that ever builds commands from a checkout, and I'd be glad to pick multi-ecosystem detection back up if it returns.

Issue #218, opened 2026-08-30
Agent triggers do work for senders without write access: refused requests still provision, and prose fails the run


cdotta:
## Summary

Two defects in the `Resolve requested agent` step region of `facility-crew.yml` and
`facility-codex.yml`. Both let a GitHub account with **`read` permission only** cause work on a
public repository that has Facility's repo lane installed.

1. A refused request still checks out the repository and installs dependencies. The job reports
   **success**.
2. A comment that merely *mentions* a command fails the run before the permission gate is
   consulted. The job reports **failure**.

The authorization gate itself is correct and does refuse the request. The problem is what
happens around it.

## 1. A refused request still provisions, and reports success

The toolchain steps rendered by `{{TOOLCHAIN_STEPS}}` carry no `if:`, while the steps on either
side of them do:

```yaml
      - name: Start agent receipt clock
        if: steps.requested-agent.outputs.run == 'true'      # guarded
        ...
      - uses: pnpm/action-setup@…                            # NOT guarded
      - uses: actions/setup-node@…                           # NOT guarded
      - run: pnpm install --frozen-lockfile                  # NOT guarded

      - name: Provision environment
        if: steps.requested-agent.outputs.run == 'true'      # guarded
```

`actions/checkout` in `facility-crew.yml` is likewise unguarded.

An account with `read` commented `/architect`. The gate refused it, and the job then did the
work anyway:

```
##[warning]Skipping crew: @cdottaaux has repository permission 'read',
          but /builder and /architect require write access.
...
Run pnpm/action-setup            → Done in 749ms
Run pnpm install --frozen-lockfile → Progress: resolved …
```

Step outcomes for that run:

```
job: success
  success  Resolve requested agent
  success  Run actions/checkout@…
  skipped  Start agent receipt clock
  success  Run pnpm/action-setup@…
  success  Run actions/setup-node@…
  success  Run pnpm install --frozen-lockfile
  skipped  Provision environment
  skipped  Run /builder
  skipped  Run /architect
```

Run: https://github.com/cdotta/remote-mcp-starter/actions/runs/33229949990

No agent runs and no model spend occurs, which is the important boundary and it holds. But any
account on the internet can trigger a full checkout, toolchain setup, and dependency install on
demand, repeatedly, and every one of those runs is green. On a repository with a real provision
command the same path would also execute that command — for a request that was explicitly
refused.

`facility-codex.yml` guards its checkout but not its toolchain block, so the same event fails
[...]

PR #221, opened 2026-08-30
fix(cli): authorize the sender before judging the command shape

cdotta:
Relates to #218.

**If this does not match how you want the triggers to behave, please just close it — no explanation needed.** It changes behavior rather than fixing something plainly broken, so the call is yours. I wrote it because I had the reproduction in front of me and it seemed more useful to show the change than describe it. The evidence stands in #218 either way.

## Problem

Both agent triggers validate command shape and exit non-zero **before** checking whether the sender has write access. In `facility-crew.yml` the error is at line 110 and the permission gate at line 160.

The job-level prefilter matches the command anywhere in a comment body, so any account on a public repository reaches that step. An account with `read` permission commenting prose produced:

```
##[error]No supported agent command found. Use /builder or /architect at the start of a line.
##[error]Process completed with exit code 1.
```

It also affects maintainers: writing about the commands in an issue turns the repository's checks red. The step's own comment says the line-anchored pattern "ignores prose examples (`ask /architect about it`)" — it does ignore them for dispatch, but the run still fails loudly rather than skipping.

`facility-codex.yml` already handles the same input the way I would expect (`::notice::No Codex command starts a line; treating the mention as prose`), so the two sibling workflows currently disagree.

## Change

The authorization gate runs first, and a body with no line-anchored command becomes a `run=false` skip rather than a hard error — matching what codex already does. A genuinely ambiguous request from an authorized sender (both commands, line-anchored) stays a hard error.

## Test

`packages/cli/test/crew-resolve.test.mjs` renders both workflows through the real installer, lifts the resolve step's `run:` block out of the rendered YAML, and executes it under bash against real event payloads with a stubbed `gh` permission lookup.

Statement order is not observable from a text assertion, which is why CONTRIBUTING's rule about exercising the rendered workflow rather than checking text matters here. 8 tests; 3 fail on `main`.

The tests require bash 4 for `mapfile`, as the shipped script does, and skip on older shells — macOS still ships bash 3.2, so a contributor there would otherwise see a false failure.

## Deliberately not included

The unguarded toolchain steps. A refused request still checks out the repository and runs `pnpm install --frozen-lockfile`, because `{{TOOLCHAIN_STEPS}}` carries no `if:` while the steps either side of it do — the job then reports success. Guarding those steps is the right fix, but `toolchainSteps()`'s existing `conditional` option guards on `steps.workflow-change.outputs.changed`, so it is not a drop-in, and the shape of that fix is yours to choose. Evidence and run logs in #218.

### DIFF
--- a/packages/cli/templates/workflows/facility-codex.yml
+++ b/packages/cli/templates/workflows/facility-codex.yml
@@ -72,15 +72,9 @@ jobs:
             [ "$mode" = builder ] && has_builder=true
             [ "$mode" = architect ] && has_architect=true
           done
-          if [ "$has_builder" = true ] && [ "$has_architect" = true ]; then
-            echo "::error::Use exactly one command: /codex-builder or /codex-architect."
-            exit 1
-          fi
-          if [ "$has_builder" = false ] && [ "$has_architect" = false ]; then
-            echo "::notice::No Codex command starts a line; treating the mention as prose."
-            echo "run=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
+          # Authorization decides before command shape does: the job prefilter matches
+          # the command anywhere in the body, so on a public repository any sender
+          # reaches this step and must leave a skip, never a red run.
           sender="$(jq -r '.sender.login // ""' "$GITHUB_EVENT_PATH")"
           permission="$(gh api "repos/$GITHUB_REPOSITORY/collaborators/$sender/permission" --jq '.permission' 2>/dev/null || echo none)"
           case "$permission" in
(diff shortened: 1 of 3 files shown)


adrian-lorenzo (changes requested):
Thanks for the contribution!

The resolver now rejects unauthorized senders before command-shape validation, but in `facility-crew` that decision still does not gate checkout or `{{TOOLCHAIN_STEPS}}`. On `pull_request_review_comment`, the default checkout uses the PR merge ref, so an outsider can post a comment containing `/builder` and make the job run `pnpm install`, `npm ci`, or `yarn install` from PR-controlled code even though the resolver writes `run=false`.

Please guard checkout and every generated setup/install step with `steps.requested-agent.outputs.run == true`, and add rendered-workflow coverage proving a rejected request cannot execute them. Once that authorization boundary is enforced, this should be ready.

javiertoledo:
Closing this one — we merged #289, a big refactor that simplifies the system. The two workflow templates this patched, `facility-crew.yml` and `facility-codex.yml`, no longer exist; vendored workflows became agent definitions under `packages/cli/templates/agents/`.

You framed this as "close it if this isn't how you want the triggers to behave", and this is a duller reason than that — the triggers themselves were rewritten. The observation still holds where it applies: authorize first, then judge the shape, and treat prose as a skip rather than an error.

Thanks for this and for #220, which is on main.

Issue #240, opened 2026-08-31
engines admits Node versions the pinned pnpm@11 cannot run on (node:sqlite), so an engines-compliant checkout dies before install


ophiocus:
## Summary

`package.json` declares `"engines": { "node": ">=22" }` and pins `"packageManager": "pnpm@11.20.0"` — but pnpm 11 hard-requires the `node:sqlite` builtin, which is **flag-gated until late in the Node 22/23 lines**. The engines range therefore admits a band of Node versions on which the pinned package manager cannot start at all, and the contributor sees a pnpm stack trace rather than anything naming the real constraint. Windows contributors, who tend to sit on nvm-windows-installed even-numbered releases, walk straight into it.

Adjacent to #167 (corepack missing on Node 25+ breaks the documented quickstart) but a different clause: #167 is about how pnpm gets installed; this is about `engines` promising Node versions the pinned pnpm cannot run on.

## Observed ladder (Windows 11, nvm-windows, clean checkout of `main`)

Every rung verbatim, same repo, same day:

1. **Node 20.18.0** — `corepack pnpm` via a stale shim: `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING` (corepack relic; #167 territory).
2. **Node 20.18.0** — `npm i -g pnpm@11 && pnpm install`: `ERR_UNKNOWN_BUILTIN_MODULE` from pnpm's own bundle (no `node:sqlite` in 20.x). Engines says >=22, fair — but the error names nothing actionable.
3. **Node 23.3.0** (satisfies `>=22`) — `npx pnpm@11 install`: `ERR_UNKNOWN_BUILTIN_MODULE: No such built-in module: node:sqlite` at `../store/index/lib/index.js`. **Engines-compliant Node, pinned pnpm, cold stop.**
4. Node 22.23.2 (WSL, same machine): everything works — which is the point: the working floor is a *late* 22.x, not ">=22".

## Why it matters

The method doc's own thesis is that agents (and people) hedge when "the environment can't run anything." A first-touch contributor on an engines-compliant Node gets a pnpm internals stack trace before reaching `pnpm install`'s first byte of output. Nothing in README, CONTRIBUTING, or the error path names `node:sqlite` or the real minimum.

## Suggested fix (small)

- Raise `engines.node` to the actual floor pnpm 11 needs (the first release line with unflagged `node:sqlite` — e.g. `>=22.13`), or document the floor in CONTRIBUTING next to the corepack note from #167.
[...]

Issue #230, opened 2026-08-31
init misreports the failed .agents/skills symlink on Windows (EPERM, not a filesystem limitation); a junction would work


Julian-Genuario:
## Summary

On Windows, `init` prints

```
! .agents/skills symlink could not be created (filesystem without symlink support) — skipped.
```

The filesystem supports symlinks; what fails is `symlinkSync(..., "dir")` with `EPERM`, because Windows only grants `SeCreateSymbolicLinkPrivilege` to administrators or with Developer Mode on. A directory junction needs no privilege and works for this purpose:

```
> node -e "fs.symlinkSync('../.claude/skills','.agents/skills','dir')"
code: EPERM | EPERM: operation not permitted, symlink '..\.claude\skills' -> '...\.agents\skills'
> node -e "fs.symlinkSync('../.claude/skills','.agents/skills','junction')"
junction OK
```

## Where

`src/init.mjs:513-519` catches every error from `symlinkSync` and reports the same message.

## Why it matters

`.agents/skills` is how non-Claude tooling discovers the vendored skills (`init.mjs:507`, `concepts/method.md:153`, `validate-delivery-loop.md:186`), so on a Windows workstation the Codex lane runs without them and the warning points the user at the wrong cause.

## Fix

On `EPERM`, retry with `type: "junction"` (Node resolves the target to an absolute path for junctions, which is fine here); if that fails too, print the actual error code instead of guessing. One-line change plus a test that asserts either a symlink or a junction exists after `init` on `win32`. Happy to open the PR.

ophiocus:
Reproduced on this machine (Windows 11, Node 22.23.2, no Developer Mode): the symlink attempt comes back EPERM and init's message points at the filesystem rather than at the actual cause and remedies.

Are you taking it? Otherwise I'd like to fold it into the Windows arc (#240→PR #246, #241→PR #245): catch EPERM/EINVAL on win32 specifically, fall back to a junction or file-copy where symlinks aren't available, and report honestly which mode was used — plus the message naming the two real fixes (Developer Mode, or an elevated shell). PR #245's verify-windows CI job currently skips the init end-to-end test with this issue named as its unskip criterion, so the fix lands with its regression test already waiting. Your find, your right of way.

Issue #228, opened 2026-08-31
init renders workflow names, check commands and the default branch into YAML/JSON/regex without escaping for that context


Julian-Genuario:
## Summary

`render.mjs` substitutes `{{VARS}}` verbatim, which is the right primitive, but three consumers put user- or repo-controlled strings into a structured context (YAML, JSON, a regex literal) without escaping for that context. Each one produces a file that is silently invalid. #217 is the same shape for `{{PROVISION_CMD}}`; these are the remaining ones I could reproduce.

### 1. `DOCTOR_WATCH`: detected workflow names rendered as bare YAML scalars

`src/detect.mjs:139` deliberately strips the quotes from `name: "..."`, and `src/init.mjs:103` writes the result back raw under `on.workflow_run.workflows` in `facility-doctor.yml`. A workflow named `CI: Build & Test` renders as `- CI: Build & Test` (a mapping — GitHub rejects it); `[nightly] e2e` fails to parse outright. Either way the doctor workflow never fires, and `workflow_run` errors are only visible in the Actions tab.

Fix: `JSON.stringify(name)` in `doctorWatch`.

### 2. `checksAllowJson`: check commands spliced into `.claude/settings.json` without JSON escaping

`src/init.mjs:113`: `` `      "Bash(${c})",` ``. A check like `node -e "process.exit(0)"` produces invalid JSON (`Expected ',' or ']' … line 7`), and Claude Code then ignores the whole settings file — permissions *and* hooks.

Fix: `JSON.stringify(`Bash(${c})`)`.

### 3. `DEFAULT_BRANCH` spliced into a regex literal in `protect-branch.mjs`

`templates/claude/hooks/protect-branch.mjs:10`: `const PROTECTED = /^(origin\/)?({{DEFAULT_BRANCH}}|main|master)$/;`. With `--branch=release/2026` the rendered file is

```js
const PROTECTED = /^(origin\/)?(release/2026|main|master)$/;
// SyntaxError: Invalid regular expression flags
```

The hook now exits 1 on every Bash call (Claude Code logs the error and lets the command through), so exactly the repositories with non-trivial branch names lose the force-push/direct-push guard. `.`, `+`, `(` in a branch name misbehave more quietly.

Fix: render the branch with `JSON.stringify` and build the regex with `new RegExp` over an escaped string.

## Test gap

[...]

PR #245, opened 2026-08-31
fix(cli): engage the gh stubs on Windows instead of calling the real API

### LINKED ISSUE #241: test(cli): suite is POSIX-only — gh stubs never engage on Windows, so eight tests call the real GitHub API
## Summary

The CLI test suite is POSIX-only in three independent ways, and the sharpest consequence is not "tests fail on Windows" — it is that **eight tests silently bypass their own `gh` stub and call the real GitHub API** from the developer's machine.

Run context: Windows 11, Node 22.23.2, pnpm 11.20.0 (stable, no flags), clean `main`, `pnpm install --frozen-lockfile` green. `node --test test/*.test.mjs` in `packages/cli`: **86 pass / 11 fail**.

## Failure class 1 — the `gh` stub never engages (8 tests, the important one)

`test/delivery.test.mjs` and `test/doctor-policy.test.mjs` stub `gh` by writing an **extensionless shell script**, `chmod 0o755`, and prepending the fixture dir to PATH **joined with a hard-coded colon**:

```js
// delivery.test.mjs:149
PATH: `${fixture.dir}:${process.env.PATH}`,
// delivery.test.mjs:140, doctor-policy.test.mjs:421
chmodSync(ghPath, 0o755);
```

On Windows the PATH delimiter is `;`, so the whole variable is corrupted rather than prepended — and even with the right delimiter, an extensionless script is not executable there. The child process therefore resolves the **real `gh`**, and the tests fire live requests:

```
gh: Not Found (HTTP 404)
Error: Command failed: gh api repos/acme/demo/pulls/7
```

Five delivery-verifier tests and the three resolver tests fail this way — including the one named *"resolver integrates with **deterministic GitHub fixtures**"*, which on this platform is hitting api.github.com. If a repo named `acme/demo` existed with a permissive API surface, these tests would read from it. A unit suite whose stub failure mode is "talk to production GitHub with the developer's credentials" is worth closing on any platform.

## Failure class 2 — POSIX permission assertions (2 tests)

`login verifies /v1/me and writes config with 0600 permissions` and `profiles can be listed and switched` assert mode `0600`; NTFS has no POSIX modes, so the assertion can never hold on Windows.

## Failure class 3 — the `.agents/skills` symlink (1 test)

`init installs the method end to end` dies on `ENOENT … .agents\skills` — this is the product-side bug already filed as #230 showing up in the suite; listed here only for completeness of the 11.

## Suggested fix (small, mechanical)

- Join PATH with `path.delimiter`, not `":"`.
- Write the stub as `gh.cmd` on Windows (or use a tiny `.mjs` invoked via a `gh.cmd`/`gh` pair) — the fixture helper can emit both in one place since both test files share the pattern.
- Guard the `0600` assertions with `process.platform !== "win32"` (or assert via `fs.stat` only where modes are meaningful).
- Optionally the deeper safety: have the stub tests set `GH_HOST`/`GH_TOKEN` to an invalid value so that *if* the stub ever fails to engage again, the suite fails fast offline instead of reaching the real API.

Happy to send the PR — I have the failing Windows environment live for verification, and it complements the Windows arc already in flight (#182, #230, #240).

ophiocus:
Closes #241. The suite's `gh` stubs never engaged on Windows — the fixture PATH was joined with a hard-coded `:` and the stub is an extensionless script — so eight tests resolved the developer's **real `gh` and called api.github.com**, including the one named *"deterministic GitHub fixtures"*. This PR makes the stubs engage on every platform, adds a belt so a future fallthrough can never reach the real API, and stands up the `verify-windows` CI job so the whole Windows-only defect class fails in CI instead of on contributors' machines.

## Why a seam, not a shim

The obvious fix — a `gh.cmd` wrapper beside the stub — cannot work: Node refuses to execute `.cmd`/`.bat` through `execFile` without a shell **by design** since CVE-2024-27980 (the same refusal behind #182's `spawn EINVAL`). No PATH arrangement makes a script stub executable there. So the templates gain a minimal, shell-free seam:

```js
const GH_BIN = process.env.FACILITY_GH_BIN ?? "gh";
const GH_ARGS = process.env.FACILITY_GH_ARGS ? JSON.parse(process.env.FACILITY_GH_ARGS) : [];
```

Production behavior is byte-identical (default `"gh"`, no args, no shell ever). Tests set `FACILITY_GH_BIN=process.execPath` and pass the stub script via `FACILITY_GH_ARGS` — works identically on POSIX and Windows.

## The changes

| What | Where |
|---|---|
| Seam in the two gh-calling templates | `templates/delivery/verify.mjs`, `templates/doctor/resolve.mjs` |
| Tests drive the seam; PATH joined with `path.delimiter` | `test/delivery.test.mjs`, `test/doctor-policy.test.mjs` |
| **Fail-offline belt**: stub env pins `GH_HOST=gh-stub.invalid` + dummy token, so if a stub ever falls through again the suite dies against an unresolvable host, never against the real API | both test files |
| POSIX `0600` asserts scoped to platforms with mode bits (NTFS has none); the secret-not-logged assert still runs everywhere | `test/platform.test.mjs` |
| init e2e skipped on win32 **with #230 as the recorded unskip criterion** (its symlink bug, not this PR's subject) | `test/init.test.mjs` |
| **`verify-windows` job**: CLI suite + guards on `windows-latest`, same pinned action SHAs as `verify` | `.github/workflows/ci.yml` |

## Why the CI job belongs here

Ten currently-open issues are Windows-only defects invisible to `ubuntu-latest` (#182 #183 #167 #191 #217 #227 #228 #230 #240 #241 — inventory in #241). Every one shipped because nothing red ran on Windows. The job covers the pure-Node surfaces (CLI suite, guards) in ~3 minutes — no Docker, no Postgres — and this very PR is its first proof: before it, that job fails 11 tests; with it, green.

## Verified

- **Windows 11, Node 22.23.2, pnpm 11.20.0** (the environment #241 was found on): 86 pass / 11 fail before → **96 pass / 0 fail / 2 skip** after.
- **Linux (WSL, same versions)**: **97 pass / 0 fail / 1 skip** — no regression; the seam's default path is exercised by the entire existing suite.
- Biome clean through the init e2e's own formatter gate over the installed templates.

Found while running the suite on a real Windows machine — happy to iterate on any of it, and the Windows arc continues in #240/#227.


adrian-lorenzo (changes requested):
Thanks for the contribution!

`FACILITY_GH_BIN` and `FACILITY_GH_ARGS` are now trusted from the ambient environment by the shipped delivery verifier. The builder and repository checks run before the delivery steps, and either can persist these variables through `$GITHUB_ENV`. The verifier would then execute the selected binary with `GH_TOKEN`, allowing earlier code to replace `gh` and fake the GitHub evidence or run inside the privileged finalization step.

Please keep the test seam at the module boundary—for example, by injecting the GitHub runner into an exported entry point—while the production entry point always invokes the fixed executable. Add a regression test showing that hostile ambient variables cannot redirect it.

The Windows CI coverage and offline safeguard are worth keeping. Once this boundary is sealed, this will be a strong improvement.

ophiocus:
Sealed in 48e510a — and thank you for spelling out the attack path; `$GITHUB_ENV` persistence from an earlier step is exactly the kind of thing I should have modeled before reaching for env vars at all.

The seam now lives where you pointed:

- **`verify.mjs`** exports `runDelivery(mode, { gh })` and **`resolve.mjs`** exports `main({ gh })` — the injectable runner exists only as a parameter on those entry points. The script paths construct the fixed `"gh"` executable inline; nothing about the invocation is readable from the environment anymore.
- The delivery tests and the resolver integration tests now inject the runner **in-process through the exported entry points** (the doctor tests got simpler for it — the fixture map became a closure instead of a stub binary).
- **Two spawn-path regression tests** prove the boundary holds: each launches the real script with a hostile `FACILITY_GH_BIN` pointing at a marker-writing binary plus `GH_HOST=gh-stub.invalid` — the marker is never written, and the fixed gh dies offline instead of reaching the API. If anyone ever reintroduces an ambient override, those two go red.

Verified: Linux 99 pass / 0 fail, Windows 98 pass / 0 fail on the same head; the `verify-windows` job you liked stays in and re-proves it on the hosted runner.

ophiocus:
Status note: `verify-windows` and `minimum-node` are green on the merged head; the `verify` red is not this PR — two `browserslist` advisories published ~18:00Z today fail the audit step of every run repo-wide. #267 carries the one-line floor (house override pattern) that clears it for all open PRs.

ophiocus:
Closing this one: #289 removed the code it fixed — the delivery and doctor templates are gone, the CLI no longer calls `gh`, and the CI matrix was rebuilt without a Windows job — so there is nothing left for these stubs to guard. Thanks for the earlier review, @adrian-lorenzo; the point about exercising the stubs rather than the real API still stands wherever a `gh` call comes back.

PR #247, opened 2026-08-31
fix(cli): make --canary-bot reachable on init

### LINKED ISSUE #229: init advertises --canary-bot but the flag validator rejects it, so the canary can never be enabled
## Summary

`init` reads `--canary-bot` (`src/init.mjs:333`) and its closing checklist tells the user to "re-run init with `--canary-bot=<your-app>[bot]`" (`src/init.mjs:563`), but `validateLocalFlags` in `src/cli.mjs` does not allowlist the flag, so the CLI exits before `init()` runs:

```
$ node bin/facility.mjs init --yes --canary-bot='my-app[bot]'
Unknown option: --canary-bot
exit=0
```

The flag is also missing from `help()`, and no test exercises it (the tests go through `runCli`, so they would have hit the validator).

## Consequence

The documented way to enable the weekly canary is unreachable. `{{CANARY_BOT}}` stays `facility-canary[bot]` in `facility-crew.yml`, and the canary "skips politely" forever, which the watchtower then reports as a healthy skip rather than a misconfiguration.

## Fix

Add `canary-bot` to the allowed set and the value-taking list for `init` in `src/cli.mjs`, mention it in help, and add a test that renders with the flag and asserts on `{{CANARY_BOT}}`. Happy to open that PR.

ignaciocanosa:
Fixes #229.

## What was broken
`init`'s closing checklist tells the user to re-run it with `--canary-bot=<your-app>[bot]` to enable the weekly canary, but the flag was missing from `validateLocalFlags`'s allowlist in `cli.mjs` — the CLI rejected it with `Unknown option: --canary-bot` before `init()` ever ran. The documented path to enable the canary was unreachable, and the watchtower reports the resulting default-skip as healthy rather than as a misconfiguration.

## Fix
Added `canary-bot` to the `init` allowlist, to the value-taking flags list (it takes a value), and to the global help text.

## Verification
- New tests in `packages/cli/test/init.test.mjs`: one runs `init` with `--canary-bot=my-app[bot]` and asserts it's rendered into `facility-crew.yml` (and that the default doesn't leak in when overridden); the other asserts a valueless `--canary-bot` still errors correctly and that `--canary-bot` is documented in global help.
- Reverted the fix locally and reran: both new tests fail with exactly `Unknown option: --canary-bot`, nothing else. Reapplied — both pass.
- Ran the full `packages/cli` suite (108 tests): no new failures. 12 tests already fail on a clean checkout of `main` on this machine (chmod-permission assertions, fixtures needing infra unavailable here) — confirmed against an unmodified checkout with the same command, unrelated to this change.
- Manually reproduced the issue's exact repro command end to end against a temp repo, before and after the fix.

### DIFF
--- a/packages/cli/src/cli.mjs
+++ b/packages/cli/src/cli.mjs
@@ -93,6 +93,7 @@ function help() {
   item(dim('            --build-model=<id> --review-model=<id> --plan-model=<id>'));
   item(dim('            --codex-build-model=<id> --codex-plan-model=<id> --org=<org> --project=<n>'));
   item(dim('            --preview-image=<image> --preview-command=<cmd> --preview-port=<n> --preview-readiness-path=</health> --preview-ttl-hours=<n>'));
+  item(dim("            --canary-bot=<your-app>[bot]"));
   item(dim("Run facility <command> --help for precise usage."));
   item(dim("Global platform flags: --profile <name> --json --timeout <seconds>"));
   console.log("");
@@ -215,6 +216,7 @@ function validateLocalFlags(command, flags) {
       "preview-port",
       "preview-readiness-path",
       "preview-ttl-hours",
+      "canary-bot",
       "help",
     ]),
     add: new Set(["dir", "help"]),
@@ -277,6 +279,7 @@ function validateLocalFlags(command, flags) {
             "preview-port",
             "preview-readiness-path",
             "preview-ttl-hours",
+            "canary-bot",
           ]
         : ["dir"];
   for (const name of valueNames) {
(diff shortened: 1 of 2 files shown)


adrian-lorenzo (changes requested):
Thanks for the contribution!

Making this flag reachable also makes its value a new input to the generated crew workflow. It is currently accepted verbatim and interpolated into both the Actions `if:` expression and a Bash comparison. A value containing quotes can therefore produce executable shell syntax in `facility-crew.yml`.

Please validate it as a supported GitHub App bot login, or safely encode it for every output context, and add regression tests covering quotes, newlines, and other hostile values. Because this workflow runs agents with elevated permissions, the unsafe input case needs to be covered before merge.

With that fixed, this should be ready.

PR #249, opened 2026-09-01
fix(cli): escape hostile values in generated init artifacts

### LINKED ISSUE #228: init renders workflow names, check commands and the default branch into YAML/JSON/regex without escaping for that context

o-mid:
Fixes #228.

## Summary
- Quote workflow names in the doctor watch list so YAML-significant titles (e.g. `CI: Build`) parse correctly.
- JSON-encode check allowlist entries in `.claude/settings.json` so shell commands with quotes stay valid JSON.
- Escape the default branch before embedding it in `protect-branch.mjs` regexes (e.g. `release/2026`).

Mirrors the escaping in both `packages/cli/src/init.mjs` and `packages/core/src/render.ts`.

## Test plan
- [x] `pnpm --filter @theagilemonkeys/facility test`
- [x] `pnpm --filter @facility/core test`
- [x] New hostile-input init regression test parses generated JSON/YAML/regex output


adrian-lorenzo (changes requested):
Thanks for the contribution!

The new encoders fix the specific YAML, JSON, and regex cases, but the default branch is still interpolated directly into privileged workflow shell source. `$(id)` is a valid Git branch name, and generating the workflows with it produces `git show "origin/$(id):..."` and `?ref=$(id)`, both of which execute command substitution.

Please treat the branch as data in every generated workflow: encode YAML fields appropriately and pass it into shell steps through an environment variable rather than embedding it in command text. Apply the fix to both renderers and add a regression using hostile but valid refs such as `$(id)` and `foo"bar`. The import-order formatting failure also needs correcting.

Once every output context is covered, this will be a strong security fix.

o-mid:
Addressed review: default branch is now passed through `FACILITY_DEFAULT_BRANCH` / `DEFAULT_BRANCH` env vars in workflow steps instead of embedding refs in shell command text. Added `DEFAULT_BRANCH_YAML` for quoted YAML scalars. Regression covers `$(id)` and `foo"bar`. Fixed import order in render.ts.

o-mid:
This predates 0.12. Facility no longer generates GitHub workflows, Claude settings, or protect-branch hooks, so the original vectors and the env-var follow-up don't apply. The remaining model/command YAML quoting is in #340.

I'd close this in favor of that PR rather than rebase a dead template set.

PR #259, opened 2026-09-01
fix(github): bind /builder to the plan the approver selected

### LINKED ISSUE #233: GitHub /builder approves the newest plan on the issue, not the one the approver read
## Summary

A `/builder` comment on GitHub is not bound to a specific plan. `githubPlanAcceptance` (`services/api/src/github/router.ts:641-674`) picks the newest `plan_acceptance` proposal on the issue whose Architect run succeeded, and the approval is written against whatever came back:

```ts
.where(and(... eq(actionTypes.name, "plan_acceptance"), eq(runs.status, "succeeded"), ...))
.orderBy(desc(proposals.createdAt))
.limit(1)
```

`COMMAND_RE` (`router.ts:60-61`) carries nothing that could disambiguate, and nothing supersedes or expires an older open proposal when a newer Architect run succeeds on the same issue. The API lane is proposal-id-bound (`POST /v1/proposals/:id/decide`), so this is GitHub-only.

## Scenario

A reviewer reads plan P1 (proposal A, published with the `facility:architect-plan:<run>:<proposalA>` marker). A teammate re-runs `/architect`; run 2 succeeds and opens proposal B seconds before the reviewer's `/builder` webhook is processed — or the reviewer simply never noticed the second publication comment. Facility approves B, dispatches the Builder on P2, and the audit trail records that the reviewer approved P2 with a valid plan hash and fresh evidence. Every downstream check in `builder-plan-policy.ts` passes, because they verify that *the approved proposal* is consistent, not that it is the one the human read.

## Why it matters

Gate 1 is the product invariant ("a person accepts the plan"). Here the person accepted *a* plan; which one is decided by ordering.

## Suggested fix

Either require the proposal id (or a plan-hash prefix) in the command — `/builder <proposalA>` — with the bare form accepted only when exactly one live proposal exists on the issue; or supersede older open `plan_acceptance` proposals when a new Architect run succeeds on the same issue, and refuse `/builder` while more than one is live. Both keep the current UX for the common case.

Adjacent but different from #222/#223 (comments invalidating an approved plan) and #147 (base commit not recorded).

islahuddinn:
## Summary

- Supersedes older open `plan_acceptance` proposals when a newer Architect run opens on the same issue, so only one live plan remains in the common case.
- Refuses bare `/builder` when multiple live plans exist on an issue (`builder_plan_ambiguous`).
- Accepts `/builder <proposal-id>` (and `/codex-builder <proposal-id>`) to bind Gate 1 approval to the plan the human actually reviewed.
- Updates Architect plan publication comments to show the explicit approval command.

Fixes #233

## Test plan

- [x] `vitest run test/plan-acceptance.test.ts test/builder-plan-policy.test.ts test/github.test.ts test/github-run-progress.test.ts`
- [ ] `vitest run test/builder-plan-policy.integration.test.ts` (requires Postgres at `DATABASE_URL`)
- [ ] Manual: run `/architect` twice on one issue, verify bare `/builder` is blocked and `/builder prop_…` approves the intended plan

Made with [Cursor](https://cursor.com)

### DIFF
--- a/services/api/src/builder-plan-policy.ts
+++ b/services/api/src/builder-plan-policy.ts
@@ -22,6 +22,8 @@ export type BuilderPlanDenialCode =
   | "builder_plan_context_invalid"
   | "builder_plan_expired"
   | "builder_plan_rejected"
+  | "builder_plan_superseded"
+  | "builder_plan_ambiguous"
   | "builder_plan_already_consumed"
   | "builder_plan_stale"
   | "builder_plan_freshness_unavailable";
@@ -31,6 +33,8 @@ const BUILDER_PLAN_DENIAL_CODES = new Set<BuilderPlanDenialCode>([
   "builder_plan_context_invalid",
   "builder_plan_expired",
   "builder_plan_rejected",
+  "builder_plan_superseded",
+  "builder_plan_ambiguous",
   "builder_plan_already_consumed",
   "builder_plan_stale",
   "builder_plan_freshness_unavailable",
@@ -427,6 +431,9 @@ async function validatePlanAcceptance(
   }
   if (proposal.state === "rejected") {
     return invalid("builder_plan_rejected", "proposal_rejected");
+  }
+  if (proposal.state === "cancelled") {
+    return invalid("builder_plan_superseded", "proposal_superseded");
   }
   if (
     (proposal.state === "open" && proposal.expiresAt.getTime() <= Date.now()) ||
--- a/services/api/src/github/plan-acceptance.ts
+++ b/services/api/src/github/plan-acceptance.ts
@@ -1,0 +1,279 @@
+import {
+  actionTypes,
+  type FacilityDb,
+  proposalEvents,
+  proposals,
+  runs,
+} from "@facility/db";
+import { and, desc, eq, gt, ne, sql } from "drizzle-orm";
+
+export const PLAN_ACCEPTANCE_PROPOSAL_ID_RE = /^prop_[0-9a-f]+$/i;
+
+export type GithubPlanAcceptanceIssueScope = {
+  orgId: string;
+  projectId: string;
+  owner: string;
+  repo: string;
+  issueNumber: number;
+};
+
+export type GithubPlanAcceptanceRow = {
+  proposal: typeof proposals.$inferSelect;
+  architectRun: typeof runs.$inferSelect;
+};
+
+export type GithubPlanAcceptanceResolution =
+  | ({ status: "resolved" } & GithubPlanAcceptanceRow)
+  | { status: "ambiguous"; liveProposalIds: string[] };
+
+export type GithubPlanAcceptanceCreateInput = {
+  orgId: string;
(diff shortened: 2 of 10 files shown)


ALVZ93:
I think there is still an atomicity window in `ensureArchitectPlanAcceptance` worth closing before merge.

Right now `supersedeOpenGithubPlanAcceptances(...)` runs **before** the new `plan_acceptance` proposal is inserted. That creates two failure/race cases:

1. If the subsequent proposal/event insert fails, the previous valid open proposal has already been cancelled, leaving the issue with no live plan.
2. Two Architect completions can both run the supersede query before either inserts its own proposal, then both insert successfully and leave two live proposals.

The new ambiguous `/builder` guard makes case 2 fail closed at approval time, which is good, but it means the "newer plan supersedes older plans" invariant itself is still not atomic.

Could the new proposal creation + open event + superseding of older proposals be done in one DB transaction, with the new proposal inserted first and the older rows cancelled in that same transaction? A regression test that forces the create step to fail (and asserts the previous proposal stays open) would also pin this down.

islahuddin-malindTech:
Thanks @ALVZ93 — you were right about the window.

`supersedeOpenGithubPlanAcceptances` now runs **after** the new `plan_acceptance` insert, in the **same transaction**, under an issue-scoped advisory lock (`architect-plan-issue:<org>:<repo>:<issue>`). Concurrent Architect completions serialize on that lock, so they can no longer both cancel siblings and then both insert.

The regression you asked for is in `builder-plan-policy.integration.test.ts`: creating the replacement proposal with an invalid action type fails, the transaction rolls back, and the previous live plan stays `open`. A second test covers the success path (insert first, then cancel older rows) so bare `/builder` binds to the kept proposal.

Rebased onto current `main`.

islahuddinn:
Thanks @ALVZ93 — you were right about the window.

`supersedeOpenGithubPlanAcceptances` now runs **after** the new `plan_acceptance` insert, in the **same transaction**, under an issue-scoped advisory lock (`architect-plan-issue:<org>:<repo>:<issue>`). Concurrent Architect completions serialize on that lock, so they can no longer both cancel siblings and then both insert.

The regression you asked for is in `builder-plan-policy.integration.test.ts`: creating the replacement proposal with an invalid action type fails, the transaction rolls back, and the previous live plan stays `open`. A second test covers the success path (insert first, then cancel older rows) so bare `/builder` binds to the kept proposal.

Rebased onto current `main`.

javiertoledo:

The plan-acceptance gate is gone. `builder-plan-policy.ts`, `github/plan-acceptance.ts`, `sandbox/orchestrator.ts` — none of your 10 files are on main. The model moved to story workspaces and turns, so there's no proposal for `/builder` to pick the wrong one of.

The underlying point — that an approval should bind to the thing that was approved, not to whatever is newest — is worth carrying into the new model if you're interested. `story-domain.ts` and `routes/v1/story-workspaces.ts` are where that flow lives now.

Thanks for the work.

PR #260, opened 2026-09-01
feat(runs)!: harden governed Builder delivery recovery

alvaroloes:
## What changes

- Derive a closed `workflowWrite` capability from the runner's already
  validated final change set.
- Keep ordinary repository-pinned installation tokens at `contents:write` and
  add `workflows:write` only when the delivery contains a path below
  `.github/workflows/`.
- Bind token issuance to the same active Facility organization and GitHub
  repository owner; reject foreign, suspended, or mismatched installations
  before minting.
- Keep the request schema strict and audit only permission names and repository
  identity, never credentials or changed paths.

## Why

GitHub rejects an otherwise valid signed `createCommitOnBranch` delivery when
the commit changes a workflow but the short-lived installation token contains
only `contents:write`. The runner already knows the complete verified diff, so
it can request the smallest additional capability without exposing an
arbitrary permission map to the agent.

## Verification

- [x] Behaviour verified beyond the test suite: ordinary, workflow,
  malformed-body, cross-organization, suspended-installation, owner-mismatch,
  and case-insensitive owner paths were exercised.
- [x] Documentation updated for the conditional GitHub App permission.

Executed checks:

- API GitHub platform lane: `67/67` passed;
- runner GitHub delivery: `19/19` passed;
- workflow path predicate: `9/9` passed;
- API and runner typechecks: passed;
- Biome and `git diff --check`: passed.

One unchanged runner process-lifecycle test can miss its existing 100 ms PID
timing window under load; the changed GitHub delivery suites pass independently
and no causal relation was found.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -37,6 +37,9 @@ FACILITY_RUNNER_IMAGE=facility-runner:dev
 # Two-phase rollout only: enable after every API replica supports durable
 # repository-write leases and the old replicas have been drained.
 FACILITY_REPOSITORY_WRITE_TRACKING_PROMOTION=0
+# Two-phase rollout only: enable after every worker understands immutable
+# governed Builder successors and every old worker has been drained.
+FACILITY_GOVERNED_BUILDER_RETRY_PROMOTION=0
 # URLs as resolved from inside a sandbox, not necessarily browser-facing URLs.
 SANDBOX_API_URL=http://host.docker.internal:4400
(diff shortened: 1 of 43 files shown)


adrian-lorenzo (approved):
Thanks for tightening this permission boundary!

adrian-lorenzo (approved):

PR #262, opened 2026-09-01
fix(gateway): meter OpenAI usage from the shapes the wire actually sends

### LINKED ISSUE #232: Gateway meters OpenAI traffic wrong: streamed Responses and Chat usage parse to zero, and cached tokens are charged on top of input
## Summary

Two defects in how the gateway meters OpenAI traffic. Together they mean Codex runs are recorded at zero cost (so hard budgets never accumulate), while the non-streamed Responses calls that *are* metered overcharge cached tokens.

### 1. OpenAI usage is parsed with Responses-only, top-level field names

`services/gateway/src/usage.ts:31-36` reads `usage.input_tokens` / `usage.output_tokens` / `input_tokens_details.cached_tokens`, and the streaming path (`usage.ts:127-129`) only ever looks at `parsed.usage`:

```ts
} else {
  mergeUsage(this.usage, usageFromJson("openai", parsed));
}
```

- **Chat Completions** report `usage.prompt_tokens` / `usage.completion_tokens` / `prompt_tokens_details.cached_tokens`, both non-streamed and in the final chunk the gateway asks for via `stream_options.include_usage` (`body.ts:48`). `prompt_tokens` appears nowhere in the gateway except as a request-body estimate key (`app.ts:529`).
- **Streamed Responses** (what Codex uses: `runner/src/index.ts` sets `wire_api="responses"` and streams) carry usage as `{"type":"response.completed","response":{"usage":{...}}}` — nested under `response`, never at `parsed.usage`.

The tests encode the wrong shapes: the Chat Completions stub returns `usage: { input_tokens, output_tokens }` (`services/gateway/test/gateway.test.ts:1347-1358`) and the `/responses` stub returns a non-streamed JSON body even when `stream: true` (`:1362-1370`), so the real SSE frame is never exercised. The Anthropic fixture (`anthropicSseBody`) mirrors the real SDK shapes, which is what makes this stand out.

**Effect.** A Codex Builder run streams through with `tee.usage` all zeros; `writeMetering` computes `cost = 0`, then `reconcileBudgetReservations(tx, reservations, 0)` (`metering.ts:141`) gives the whole pre-flight estimate back. Net spend recorded: 0; `llm_requests.input_tokens = 0`. A hard project budget of $50/day never trips for Codex; the only thing that can block a Codex run is a single request whose own estimate exceeds the remaining limit.

### 2. Cached tokens are charged on top of input for OpenAI

`packages/core/src/pricing.ts:60-64` charges `inputTokens·input + cacheReadTokens·cacheRead`. That is right for Anthropic, whose `input_tokens` excludes `cache_read_input_tokens`. OpenAI's `input_tokens` (Responses) and `prompt_tokens` (Chat) *include* `cached_tokens`; the correct charge is `(input − cached)·input + cached·cacheRead`. The table's `cacheRead` rates are discounts (e.g. `gpt-5.5` input 10 / cacheRead 1), so the intent is clearly "discount the cached slice", not "surcharge on top".

Example: gpt-5.5, 100k prompt tokens of which 90k cached. Real ≈ 10k·$10/M + 90k·$1/M = $0.19; gateway books 100k·$10/M + 90k·$1/M = $1.09 (≈5.7×). Today this bites non-streamed `/responses` calls; once (1) is fixed it would bite all Codex traffic, so both should land together.

## Suggested fix

- `usageFromJson("openai", …)`: accept `prompt_tokens`/`completion_tokens`/`prompt_tokens_details.cached_tokens` as aliases; in the SSE path also look at `parsed.response?.usage` for `response.completed`.
- `costCents`: take a provider (or a `cachedTokensIncludedInInput` flag) and subtract cached from input for OpenAI.
- Replace the two stubs with real-shaped fixtures (streamed `response.completed`, Chat `prompt_tokens`) and add a Codex-lane test that asserts a hard budget trips.

Not overlapping with #208 (custom models in the pricing table); this is about the models that *are* priced.

ophiocus:
Closes #232 — and full credit to @Julian-Genuario, whose diagnosis was exact down to the sentence "the tests encode the wrong shapes." I claimed this in the issue thread with right of way offered; the courtesy window has passed, so here is the fix, built the way #181 was.

## Stubs first, so the tests bite (the #181 pattern)

Both OpenAI stubs now emit what the wire actually sends — Chat Completions speaks `prompt_tokens`/`completion_tokens` (with `prompt_tokens_details.cached_tokens` on the final include_usage chunk), and the `/responses` stub streams the real envelope: usage nested inside `{"type":"response.completed","response":{...}}`, never at the frame's top level. **With the corrected stubs alone, tests 3 and 3b fail against the old parser exactly as the issue predicted:**

```
AssertionError: expected +0 to be 600000   // Chat streaming metered to zero
AssertionError: expected +0 to be 750000   // Codex/Responses streaming metered to zero
```

## Then the parser

- `usageFromJson` accepts both naming schemes; the stream path also reads the `response.completed` envelope (the wire Codex speaks) alongside top-level usage.
- **The double-charge is fixed at its root:** OpenAI reports input INCLUSIVE of cached tokens while `costCents` sums buckets additively in the Anthropic convention — so cached tokens are subtracted from input, or they get billed at both the input and cache-read rates.

## Pinned money math

The assertions now check the full buckets, not just presence: 1M prompt with 400k cached + 1M completion on gpt-5.5-mini = **216¢** (was silently 225 with cached double-billed and names lucky-matched); the streamed Codex path (1M/250k cached/200k out on gpt-5.6-sol) = **987.5¢** — previously **zero**, which is the hard-budget bypass in the issue title.

Gateway suite: 57 tests, 0 failures, tsc clean. Adjacent, not overlapping: #248/#251/#255 handle *incomplete* usage reconciliation; this handles usage that arrives complete but was unparseable.

### DIFF
--- a/services/gateway/src/usage.ts
+++ b/services/gateway/src/usage.ts
@@ -28,10 +28,23 @@ export function usageFromJson(provider: Provider, body: unknown): Partial<Usage>
       cacheWriteTokens: numberValue(row.cache_creation_input_tokens),
     };
   }
+  // OpenAI's two APIs name usage differently: Responses reports
+  // input_tokens/output_tokens (details under input_tokens_details), Chat
+  // Completions prompt_tokens/completion_tokens (prompt_tokens_details).
+  // Both report input INCLUSIVE of cached tokens, while costCents sums the
+  // buckets additively in the Anthropic convention - so cached tokens are
+  // subtracted from input here, or they would be billed twice.
+  const inputInclusive = numberValue(row.input_tokens) ?? numberValue(row.prompt_tokens);
+  const cacheRead =
+    nestedNumber(row.input_tokens_details, "cached_tokens") ??
+    nestedNumber(row.prompt_tokens_details, "cached_tokens");
   return {
-    inputTokens: numberValue(row.input_tokens),
-    outputTokens: numberValue(row.output_tokens),
-    cacheReadTokens: nestedNumber(row.input_tokens_details, "cached_tokens"),
+    inputTokens:
+      inputInclusive !== undefined && cacheRead !== undefined
+        ? Math.max(0, inputInclusive - cacheRead)
+        : inputInclusive,
+    outputTokens: numberValue(row.output_tokens) ?? numberValue(row.completion_tokens),
+    cacheReadTokens: cacheRead,
     cacheWriteTokens: nestedNumber(row.input_tokens_details, "cache_creation_tokens"),
   };
 }
@@ -125,7 +138,14 @@ export class UsageTee extends Transform {
           mergeUsage(this.usage, usageFromJson("anthropic", source));
         }
       } else {
-        mergeUsage(this.usage, usageFromJson("openai", parsed));
+        // OpenAI streamed shapes: Chat Completions puts usage on the final
+        // chunk's top level (requested via stream_options.include_usage),
+        // while the Responses API - the wire Codex speaks - nests it inside
+        // the response.completed envelope ({"type":"response.completed",
+        // "response":{"usage":{...}}}) and never at the frame's top level.
+        for (const source of [parsed, (parsed as { response?: unknown })?.response]) {
+          mergeUsage(this.usage, usageFromJson("openai", source));
+        }
       }
     } catch {
       // Provider bytes are still passed through; malformed telemetry frames only affect metering.
--- a/services/gateway/test/gateway.test.ts
+++ b/services/gateway/test/gateway.test.ts
@@ -502,7 +502,13 @@ describe("gateway", async () => {
     const row = (
       await db.select().from(llmRequests).where(eq(llmRequests.virtualKeyId, setup.keyId))
     )[0];
-    expect(row?.costCents).toBe(225);
+    // Chat reports prompt/completion names, with prompt_tokens INCLUSIVE of
+    // cached: 1M prompt (400k cached) + 1M completion on gpt-5.5-mini is
+    // 0.6*0.25 + 1*2 + 0.4*0.025 = $2.16 - not input-rate on the full million.
+    expect(row?.inputTokens).toBe(600_000);
+    expect(row?.cacheRead).toBe(400_000);
+    expect(row?.outputTokens).toBe(1_000_000);
+    expect(row?.costCents).toBe(216);
   });
 
   it("3b. OpenAI Responses streaming preserves the native request shape", async () => {
@@ -513,7 +519,7 @@ describe("gateway", async () => {
       input: "hello",
     });
     expect(response.status).toBe(200);
-    await response.json();
+    expect(await response.text()).toContain("response.completed");
     await waitForRequestCount(1);
     expect(stubState.lastOpenAiRequest).toMatchObject({
       model: "gpt-5.6-sol",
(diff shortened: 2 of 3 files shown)


javiertoledo:
Reviewed this and pushed one commit to the branch (`maintainer_can_modify`): `bf57d4d`, unit coverage only. Your `usage.ts` is untouched — I went looking for a reason to disagree with the approach and did not find one.

**What I verified independently**, driving the old and new parser directly rather than trusting the description:

| | `main` | this branch |
|---|---|---|
| Chat Completions streaming | `{input:0, output:0, cacheRead:0}` | 216¢ |
| Responses streaming (`response.completed`) | `{input:0, output:0, cacheRead:0}` | 987.5¢ |
| Cached at input rate | 1112.50¢ | 987.50¢ |

Both streaming paths meter to exactly zero on `main`, so "hard budgets never accumulate for Codex runs" is precise, not rhetorical. I also confirmed `costCents` sums the four buckets additively — so the subtraction is right — and that every OpenAI model in `MODEL_PRICES_USD_PER_1M` carries a `cacheRead` price, which was the obvious risk of subtracting: cached tokens do not silently become free. Your 216¢ and 987.5¢ both reproduce exactly.

**The one gap: no unit coverage.** `usage.test.ts` is the unit suite for this module and covers Anthropic thoroughly — five cases, including frames split across chunk boundaries and malformed ones — and nothing for OpenAI. `CLAUDE.md` asks for unit *and* deterministic integration tests on a change to billing.

That gap is worse than the rule alone implies. The integration test that does cover these shapes needs Postgres and **skips silently when it is unreachable**: a local run here reported `4 passed` with 32 tests skipped and exit 0. On a developer machine without the database, nothing exercised this at all.

So `bf57d4d` adds three cases, one per shape you taught the parser — Chat Completions naming with cached subtraction, the `response.completed` envelope (with the preceding `response.created` frame proving it does not erase the total), and a non-streamed body using the chat naming. All three fail against the previous parser with `inputTokens: 0`; the five Anthropic cases are untouched and still pass. `usage.test.ts` needs no database, so these run everywhere.

Gateway suite 28 passed / 1 skipped, `tsc --noEmit`, `biome check` and `node guards/run.mjs` all clean.

Two things I looked at and deliberately left alone: `cacheWriteTokens` reads only `input_tokens_details` while `cacheRead` reads both naming schemes — but neither OpenAI API reports a cache-creation count, so no wire shape reaches it, and the line predates this branch. And `Math.max(0, …)` clamps silently if cached ever exceeded input, which no observed shape produces. Neither is worth a commit.

ophiocus:
Thank you — this is the kind of review a branch dreams about: an independent verification table, a commit that closes a real gap, and explicit notes on what you chose not to touch.

The unit-coverage point lands especially well. The silently-skipping integration suite is a masked failure of exactly the class this fix targets: `4 passed` with 32 skipped reads as green on any machine without Postgres — the same shape as `main` metering streams to zero while looking alive. Putting the three wire shapes into `usage.test.ts` where they run everywhere is the right fix for that, and pairing each with the frame that could erase it (`response.created` before `response.completed`) is a nice touch.

On the two things you left alone, I read them the same way: no OpenAI wire shape reports a cache-creation count, so the asymmetric naming on `cacheWriteTokens` has nothing to receive; and the `Math.max` clamp can only fire on a shape no observed API emits. If you ever want that clamp loud instead of silent, a debug-level log inside it would make a future impossible-shape visible without changing behavior — but I agree neither belongs in this PR.

From my side the branch is exactly as you left it — `b8fe764` + `bf57d4d` — and good to go.

javiertoledo:
Thanks — and agreed on leaving both out.

On the clamp: I'd go one step further and say a debug log is the version I'd argue against, precisely for the reason you spotted. `4 passed` with 32 skipped and `main` metering a live stream to zero are the same failure — something reporting health while measuring nothing — and a debug line inherits that shape rather than breaking it. It fires on a branch nothing reaches, into a level nobody watches, so the first time it ever mattered it would already have scrolled past.

If that state is worth knowing about, the useful version is louder and lives elsewhere: metered usage that cannot be reconciled against what the provider reported is a signal for whatever already watches spend, not a line in the parser. That is a design question rather than a line change, and it belongs to whoever owns budget reconciliation — adjacent to the #248/#251/#255 work you flagged as out of scope here. Happy to open an issue framed that way if you want it tracked; I'd rather not file "add a debug log" as a ticket, since that is the kind that sits open forever.

Branch confirmed as you describe: `b8fe764` + `bf57d4d`, CI green across all six checks — including `verify`, which is the one that actually stands Postgres up and runs the integration cases that skip locally.

One thing worth flagging for whoever merges: `bf57d4d` is `test:` and your commit is `fix(gateway):`, so the range still releases a patch and the PR title's impact matches. Nothing to change — just so the release classification isn't a surprise.

PR #263, opened 2026-09-01
fix(init): create a junction for .agents/skills on Windows and report link failures honestly

### LINKED ISSUE #230: init misreports the failed .agents/skills symlink on Windows (EPERM, not a filesystem limitation); a junction would work

ophiocus:
Closes #230 — credit to @Julian-Genuario for the precise report: EPERM on Windows without Developer Mode, blamed on "filesystem without symlink support." Claimed in-thread with right of way offered; the courtesy window has passed, so here is the fix, verified on the reporting platform.

## The fix

- **Directory junctions need no privileges on Windows**, so on EPERM/EINVAL init now falls back to one (absolute target — the ok line says so honestly: *"junction — absolute; recreate after moving the repo"*) instead of failing.
- When even the junction fails, the warn names the **two real remedies** — Developer Mode, or an elevated shell — instead of misdiagnosing the filesystem.
- Every other platform keeps the original relative symlink path, byte-identical.

## Found underneath (masked until now)

The init end-to-end test also spawned Biome through its POSIX `.bin` shim, which `spawnSync` cannot execute on Windows (status `null`) — invisible before because the symlink death always came first. It now runs Biome's JS entry through node, same behavior everywhere.

## Verified

`init.test.mjs`: **8/8 on Windows 11 — the first time that file has ever passed there** — and 8/8 on Linux, unchanged.

Interplay with #245, stated plainly: that PR's `verify-windows` CI job currently skips the init e2e naming this issue as its unskip criterion. Once both land, a one-line follow-up removes the skip and the whole init flow is guarded on `windows-latest` forever. Happy to push that follow-up the moment the second of the two merges.

### DIFF
--- a/packages/cli/src/init.mjs
+++ b/packages/cli/src/init.mjs
@@ -6,7 +6,7 @@
 import { chmodSync, existsSync, lstatSync, mkdirSync, readFileSync, symlinkSync, writeFileSync } from "node:fs";
 import { createHash } from "node:crypto";
 import { pathToFileURL } from "node:url";
-import { dirname, join } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { detect } from "./detect.mjs";
 import { render, hasManagedBlock, appendManagedBlock } from "./render.mjs";
 import { ask, confirm, closePrompts } from "./prompts.mjs";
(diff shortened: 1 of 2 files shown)


ophiocus:
Closing this one: #289 rewrote `init` and it no longer creates the `.agents/skills` link at all — skills are inventoried from the repository instead — so the Windows junction has nothing to attach to. Happy to revisit if a link path comes back.

PR #264, opened 2026-09-01
fix(doctor): notice receipts destroyed after run.finished was audited

### LINKED ISSUE #226: receipt_integrity never notices a receipt that disappears after run.finished was audited
## Summary

`verifyStoredReceipts()` only walks runs that still have a receipt and checks each one against the `run.finished` digest in the audit log. It never asks the reverse question: is there a `run.finished` event whose `receipt_sha256` no longer has a receipt behind it? A receipt that is nulled or a run row that is deleted after the fact simply drops out of the check, `ok` stays `true`, and the doctor's "N stored agent receipts verify against the audit chain" quietly reports a smaller N.

## Where

`services/api/src/receipt-integrity.ts:20-29` selects `where receipt is not null`; `:34-41` builds `auditedDigests` from the audit log, but `:44-53` only consults that map for rows that came back from the first query.

In current code this is always an anomaly: every terminal run gets a non-null receipt in the same UPDATE as its status (`services/api/src/sandbox/orchestrator.ts:565-572`) and then a `run.finished` event carrying its digest (`:704-715`). "Audited digest exists, receipt missing" can only mean the receipt was destroyed, the run row was deleted, or a migration nulled it.

## Failure scenario

```sql
update runs set receipt = null where id = 'run_x';   -- or delete from runs where id = 'run_x'
```

after a builder run that Gate 1 or a learning packet relied on. `facility doctor` → `receipt_integrity: pass`, `checked` is one lower, nothing distinguishes "never had a receipt" from "receipt destroyed". `learning.ts:228` will show `present: false` for that run only if it falls inside a packet window, and that is not treated as a failure.

## Coverage

The only test of `verifyStoredReceipts` is the positive case at `services/api/test/sandbox.test.ts:1609`; no test ever produces `invalidRunIds` or `unauditedRunIds`.

## Suggested fix

Iterate `auditedDigests` after the existing loop and put every run id that is absent from `stored` into a new `missingReceiptRunIds` bucket that fails the check; surface it in the doctor message alongside the existing two. Add negative tests for all three buckets.

ophiocus:
Closes #226 — credit to @Julian-Genuario for spotting the missing reverse question. Claimed in-thread with right of way offered; the courtesy window has passed, so here is the fix in the shape the claim promised.

## The gap, closed

`verifyStoredReceipts()` proved every *surviving* receipt against the audit chain but never asked whether an audited `run.finished` digest still has a receipt behind it — so `UPDATE runs SET receipt = null` (or deleting the row) just shrank `checked` while `ok` stayed `true`, exactly as the issue demonstrated.

- The reverse pass now runs over the audited digests: any id with an audited digest and no stored receipt lands in a **new, distinct `missingReceiptRunIds` class** — kept separate from `invalidRunIds` and `unauditedRunIds` so "receipt destroyed" and "never audited" remain distinguishable stories, per the issue's point that in current code this state can only be an anomaly.
- The `runIds` scope is respected (asking about other runs stays clean), and the doctor's failure line now reports all three counts.

## The regression bites

The test finishes a real run through `/internal/runs/:id/result`, confirms `ok: true`, then nulls the receipt — doctor goes red with `missingReceiptRunIds: [run.id]` — and finally deletes the run row entirely, which trips the same wire because the audit trail outlives it. The "absence must fail loudly" shape, same family as the truncation disclosure above it in the suite.

`sandbox.test.ts`: 40 tests, 0 failures; tsc clean.

### DIFF
--- a/services/api/src/doctor.ts
+++ b/services/api/src/doctor.ts
@@ -740,7 +740,7 @@ async function checkReceiptIntegrity(db: Db, orgId: string): Promise<DoctorCheck
     return fail(
       "receipt_integrity",
       "Agent receipt integrity",
-      `${report.invalidRunIds.length} invalid and ${report.unauditedRunIds.length} unaudited receipts found across ${report.checked} runs.`,
+      `${report.invalidRunIds.length} invalid, ${report.unauditedRunIds.length} unaudited, and ${report.missingReceiptRunIds.length} audited-but-missing receipts found across ${report.checked} runs.`,
       "Stop outcome and learning jobs, preserve audit_events and runs, then investigate receipt mutation or an outdated runner.",
     );
   }
--- a/services/api/src/doctor.ts
+++ b/services/api/src/doctor.ts
@@ -740,7 +740,7 @@ async function checkReceiptIntegrity(db: Db, orgId: string): Promise<DoctorCheck
     return fail(
       "receipt_integrity",
       "Agent receipt integrity",
-      `${report.invalidRunIds.length} invalid, ${report.unauditedRunIds.length} unaudited, and ${report.missingReceiptRunIds.length} audited-but-missing receipts found across ${report.checked} runs.`,
+      `${report.invalidRunIds.length} invalid, ${report.unauditedRunIds.length} unaudited, and ${report.missingReceiptRunIds.length} audited-but-missing receipts found across ${report.checked + report.missingReceiptRunIds.length} audited runs.`,
       "Stop outcome and learning jobs, preserve audit_events and runs, then investigate receipt mutation or an outdated runner.",
     );
   }
--- a/services/api/src/receipt-integrity.ts
+++ b/services/api/src/receipt-integrity.ts
@@ -7,6 +7,12 @@ export type ReceiptIntegrityReport = {
   checked: number;
   invalidRunIds: string[];
   unauditedRunIds: string[];
+  /** Audited run.finished digests whose run no longer carries a receipt —
+   * the receipt was destroyed or the run row deleted after the fact. Every
+   * terminal run writes its receipt and its audit event together, so this
+   * can only be an anomaly; it is reported distinctly from "never audited"
+   * so the two failure stories stay distinguishable. */
+  missingReceiptRunIds: string[];
 };
 
 export async function verifyStoredReceipts(
@@ -15,7 +21,7 @@ export async function verifyStoredReceipts(
   runIds?: string[],
 ): Promise<ReceiptIntegrityReport> {
   if (runIds?.length === 0) {
-    return { ok: true, checked: 0, invalidRunIds: [], unauditedRunIds: [] };
+    return { ok: true, checked: 0, invalidRunIds: [], unauditedRunIds: [], missingReceiptRunIds: [] };
   }
   const stored = await db
     .select({ id: runs.id, receipt: runs.receipt })
@@ -41,6 +47,11 @@ export async function verifyStoredReceipts(
   }
   const invalidRunIds: string[] = [];
   const unauditedRunIds: string[] = [];
+  const storedIds = new Set(stored.map((row) => row.id));
+  const scope = runIds ? new Set(runIds) : null;
+  const missingReceiptRunIds = [...auditedDigests.keys()].filter(
+    (id) => !storedIds.has(id) && (!scope || scope.has(id)),
+  );
   for (const row of stored) {
     const parsed = FacilityReceiptSchema.safeParse(row.receipt);
     if (!parsed.success || !verifyFacilityReceipt(parsed.data)) {
@@ -52,10 +63,14 @@ export async function verifyStoredReceipts(
     }
   }
   return {
-    ok: invalidRunIds.length === 0 && unauditedRunIds.length === 0,
+    ok:
+      invalidRunIds.length === 0 &&
+      unauditedRunIds.length === 0 &&
+      missingReceiptRunIds.length === 0,
     checked: stored.length,
     invalidRunIds,
     unauditedRunIds,
+    missingReceiptRunIds,
   };
 }
 
--- a/services/api/src/receipt-integrity.ts
+++ b/services/api/src/receipt-integrity.ts
@@ -21,7 +21,13 @@ export async function verifyStoredReceipts(
   runIds?: string[],
 ): Promise<ReceiptIntegrityReport> {
   if (runIds?.length === 0) {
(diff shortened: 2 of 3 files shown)


javiertoledo:
Reviewed this end to end and pushed nothing — I could not find anything that warranted a commit.

**Verified on `main` rather than taken from the description.** The query filters `receipt IS NOT NULL` and the loop walks only survivors, while `auditedDigests` is built but read solely through `.get(row.id)` — never traversed in reverse. So nulling a receipt shrinks `checked` and leaves `ok` true, exactly as #226 said.

I also went looking for the obvious way this fix could be wrong — a legitimate deletion path that would make `missingReceiptRunIds` fire as a false positive. There is none: nothing in `services/api` or `packages/db` deletes `runs` or `auditEvents`, and there is no retention or purge job. Your comment that this "can only be an anomaly" is precise, not optimistic.

Worth adding that this restores a written decision rather than inventing a rule. `architecture.md` records that "the receipt SHA-256 detects accidental or post-production changes" (decision 19). Destruction is a post-production change the check could not see, so the reverse pass closes the gap between the doc and the code.

The test earns its place: it drives a real run through `/internal/runs/:id/result`, confirms green first, then nulls the receipt and deletes the row, so it covers both shapes of the anomaly rather than restating the implementation.

**Three things I looked at and deliberately did not act on:**

The test leaves an audited-but-missing run in the shared org, and `checkReceiptIntegrity` queries org-wide with no scope. Nothing breaks today — no test asserts `receipt_integrity`, and `test-critical.mjs` runs `api` before `sandbox` with `--fileParallelism=false`. It is a landmine for whoever first writes a doctor-level test for this check, though, so it seemed worth naming even without a trigger.

The failure message reads `N audited-but-missing receipts found across ${checked} runs`, but `checked` counts only surviving receipts, so the missing ones are not in their own denominator. Null the only receipt in an org and an operator sees "1 audited-but-missing receipts found across 0 runs." Cosmetic, and I would rather not churn a string on your branch.

`receipt-integrity.ts` has no unit-test home, and `sandbox.test.ts` skips silently when Postgres is unreachable. I raised the equivalent on #262 and added the cases there — but that file already existed and covered the sibling provider, so it was three cheap cases. Here it would mean a new file plus drizzle fakes for a function whose whole job is two queries, which is more machinery than the gap justifies. `test-critical.mjs` also refuses to let a skip pass as green, so CI is not exposed.

The unbounded row loads are real but pre-existing and unchanged by this branch, so they are tracked separately in #295 rather than held against you here.

CI is green across all six checks, `verify` included. From my side this is good to merge as it stands.

ophiocus:
Thank you — a review that verifies the bug on `main`, hunts for the false-positive path and names why there isn't one, and then ties the fix back to `architecture.md` decision 19 is a review that teaches. "Restores a written decision rather than inventing a rule" is exactly the framing I was reaching for and didn't find.

On your three deliberate non-actions:

- **The denominator** — you were right, and since the branch is still open I folded it in rather than leave a known-wrong string for an operator's worst morning: `4ac067e` (on top of your main merge) divides by `checked + missing`, so it reads "found across N audited runs". One line, no other churn.
- **The org-scope landmine** for a future doctor-level test: agreed, and worth naming now. Whoever writes that test will want a scoped fixture org — I'll carry that thought into the test work I have open elsewhere if it lands on me.
- **The unit-test-home cost/benefit** — I read it the same way: two queries don't justify drizzle fakes, and `test-critical.mjs` refusing to count a skip as green is what actually protects CI here.

And thank you for routing the pre-existing unbounded loads to #295 instead of holding them against this branch — and for #262 this morning. From my side this stays good to merge.

javiertoledo:
Closing this one — #289 just landed and it deleted the code this PR fixes.

It was a big restructure, 600 files. These are all gone from main now:

```
services/api/src/receipt-integrity.ts
services/api/src/doctor.ts
services/api/test/sandbox.test.ts
```

`services/api` went from 155 files down to 91, `doctor.ts` included. There's no rebase that makes this branch apply.

The bug was real, though — I checked it on main before the merge. Forward-only verification can't see deletion, so nulling a receipt kept `ok` true.

If you want to try it again on the new code, the closest equivalents look like `stories/evidence.ts` and `turns/git-evidence.ts`. `auditEvents` is still in the schema. One thing to check first: I couldn't find any integrity verification at all in the new tree, so it may have been dropped on purpose.

Up to you. Thanks for the original catch, and to @Julian-Genuario for spotting it in #226.

ophiocus:
Thanks for the pointers, and for leaving the call open. Being useful at this level of the architecture from outside is mostly a matter of knowing which removals were decisions — without that, anything proactive is a guess that costs you more to review than it saves. Two file names and one thing to check first made it possible to look properly, so I did, as an exercise.

The check first: against `3e3e998` and the #289 description, the removal was deliberate and complete, and having mapped both trees I think it was the right call. 0.11 put trust at every operation — leases, plan policy, HITL approval, self-hashed receipts reconciled against the audit ledger — and that mediation layer is where the hard defects lived; #195, #233 and #226 were all bugs in the machinery, not in the work it guarded. 0.12 puts the boundary where it is structural and lets evidence serve delivery truth instead of adjudication. Retiring the machinery retires that class of bug with it.

What I found is smaller than the integrity verification 0.11 attempted and #264 was completing. #289 describes two boundary points per turn — *"Each turn records agent, engine, model, session, workspace, branch, and initial Git SHA before engine execution. Final SHA, commits, changed files, and dirty state are captured when it settles, including failure and cancellation paths."* — and at each of them the record is already written twice: at the start, the `turnGitEvidence` row's initial fields and the `turn:{id}:context` event; at settlement, the row's final fields and the `turn:{id}:git` event, or on the failure path a `captureError` and a `git.capture_failed` event under the same key. When work reaches GitHub there is a third copy from outside: the head SHA the mirror stored from the signed webhook, which `linkedStory()` already uses to bind a pull request to its turn by content-address. The check reads those copies against each other — row against event at both ends, and when they disagree on the final SHA, both against GitHub's head. Nothing is written, nothing is re-proven, no schema changes, and the two files you named are untouched.

Two pull requests came out of it. #308 is the read above, finished: one module and its tests, nothing else. #309 is the same shape of read applied to the table you mentioned: `auditEvents` survived the restructure as a plain time-indexed append log — no ordering, no way to tell a gap from a quiet hour — so it reads the log against the state it narrates — every API key against its `key.issued`, every `revokedAt` against its `key.revoked`, archived projects and budgets likewise — in both directions, and reports where the two disagree. Whether the log was meant to stay that way is the first thing that PR asks rather than assumes.

PR #304, opened 2026-09-04
feat(turns): warn when active stories touch the same files

### LINKED ISSUE #303: Warn agents when two active stories touch the same files
## The problem

A story is strictly serial inside itself (one queued or running turn), but stories are parallel to each other. Two agents working on two stories in the same project can both edit `src/billing/invoice.ts` on their own branches, and nothing tells either of them. The overlap surfaces days later as a merge conflict on the second pull request, after review time has been spent on both, and the person resolving it is usually not the one who understood either change.

The information needed to warn earlier already exists. `turn_git_evidence.changed_files` records every tracked and untracked path a turn changed, per story, with the branch and SHAs. Today it is only read back for the timeline; nothing compares one story's evidence with another's.

## What you propose

An advisory cross-story collision check, run once per turn between environment preparation and engine start:

1. Take the union of `changed_files` paths from this story's completed evidence rows.
2. Compare it with the same union for every other story in the project that is still active: status in `ready`, `working`, `attention`, `review`, not deleted or archived, with a branch, and whose branch does not belong to a merged pull request in the mirror.
3. For each overlapping story, append one `story.collision_detected` fact to the timeline (external key `turn:<turnId>:collision:<otherStoryId>`, so retries of the same turn are idempotent) with the other story's identity, branch, up to 20 overlapping paths and the total count.
4. Prepend a short section to the prompt listing those stories and paths, asking the agent to keep edits to them minimal and to mention the overlap in its commit or pull request description.

It is warn-only by design. It never blocks or fails a turn, does not create an attention item (attention flips the story to `attention` and clears the active agent, which is the wrong signal for a heads-up), and a detection failure is recorded as `turn.collision_check_failed` without stopping dispatch. No schema change: it uses `turn_git_evidence`, `story_evidence_events`, and the existing timeline rendering. The story `GET` and `facility_get_story` expose the facts without new fields.

What it deliberately does not do: no file locking, no per-story path reservations, no permission profiles, no change to which agent runs or when. Those would be a different product decision; this only makes the existing evidence useful before the conflict happens instead of after.

Known limit: a story with no completed evidence yet cannot collide, because its files are not known. The first turn of a new story is unwarned; every turn after that is covered.

## What you considered instead

- **Doing nothing.** Branch protection catches the conflict, but at the most expensive moment, after two reviews.
- **Reading `git diff` of the other branches at dispatch time.** Accurate, but it needs the other workspaces to be running or a fetch of every open branch on every turn. The evidence table already holds the same paths at zero runtime cost.
- **Comparing directories or top-level areas instead of exact paths.** Fewer false negatives (two files in the same module), more false positives. Exact paths are the conservative first step; a coarser mode could be added on top if it proves useful.
- **Blocking the turn or creating attention.** Rejected: an overlap is normal in a busy project, and the agent can usually handle it if told.

Open questions:

- Should merged-branch exclusion also consider `github_branches.deleted_at`, for branches that were deleted without a mirrored pull request?
- Is the prompt wording strong enough, or should the block ask the agent to add a note to the PR body explicitly?

Implementation with tests: #304

boudydegeer:
## What changes

Closes #303

Before the engine starts, the dispatcher compares the changed files recorded for the story with the recorded changed files of every other active story in the project that has an open branch. When they overlap:

- one `story.collision_detected` fact per overlapping story is appended to the timeline (external key `turn:<turnId>:collision:<otherStoryId>`, so a retried turn does not duplicate it), carrying the other story's id, title, provider identity, branch, status, up to 20 overlapping paths, and the total count;
- one `turn.collisions_detected` turn event summarizes the affected stories;
- the prompt gains a section headed "Other active stories touch files you changed" listing those stories and paths, asking the agent to keep edits to them minimal and to mention the overlap in its commit or pull request description.

Stories that are `done`, archived, deleted, without a branch, or whose branch belongs to a merged pull request in the mirror are ignored. Overlapping stories are sorted by overlap size; at most 10 are listed in the prompt and the rest are counted.

The check is advisory. It never blocks or fails a turn, never creates attention, and a detection failure is recorded as `turn.collision_check_failed` and the turn continues without the block.

The story page's timeline summary line renders the new fact as `overlaps "<title>" on N files · <branch>`.

## Why

A story is serial inside itself, but stories run in parallel, and two agents editing the same file on two branches discover it as a merge conflict on the second pull request, after both have been reviewed. `turn_git_evidence.changed_files` already holds every path each turn changed; nothing compared one story's evidence with another's. This makes that evidence useful before the conflict instead of after.

### Design notes

- **Evidence, not attention.** `flagAttention` moves the story to `attention` and clears the active agent, and an open attention item keeps the story out of `ready` after the turn. That is the right signal for "a human must act", and the wrong one for "be careful with these paths". The timeline already carries per-turn facts and is exposed through the story `GET`, the UI, and `facility_get_story` without new fields.
- **Only completed, cleanly captured evidence counts.** Rows with `capture_error` or without `completed_at` are skipped, so the current turn's own freshly inserted row and any failed capture cannot produce a false overlap.
- **A story with no completed evidence cannot collide.** Its files are unknown. The first turn of a new story is therefore unwarned; every later turn is covered. Documented in the lifecycle reference.
- **Exact paths.** Directory-level or module-level overlap would catch more and also warn more; exact paths are the conservative first step and the data model does not preclude a coarser mode later.
- **`detectCollisions` is isolated in the dispatcher** with its own try/catch so a query problem is recorded on the turn rather than turning into a failed dispatch.

## Persistence, compatibility, and release classification

- No migration and no schema change. New fact type `story.collision_detected` on `story_evidence_events` (source `facility`) and new turn event types `turn.collisions_detected` / `turn.collision_check_failed`.
- No manifest, API, SDK, or MCP contract change. The timeline entries flow through existing shapes.
- Cost: three bounded queries per dispatch, all on existing indexes (`stories_org_project_status_idx`, `github_pull_requests_project_state_idx`, `turn_git_evidence_story_completed_idx`). Only the `path` column is extracted from the JSON (`jsonb_path_query_array`), so the control plane never loads full evidence rows. Runs once per turn, before the engine; the engine's own runtime dominates by orders of magnitude.
- Prompt size: at most 10 stories × 20 paths plus two sentences, well under the existing transcript budget.
- No cost, budget, credential, or security implication. The prompt lists file paths that the same principal can already read from the timeline of every story in the project.
- Suggested title classification: `feat` (patch in 0.x). Not breaking: existing stories, manifests, and clients are unaffected.

## Verification

- [x] `pnpm verify` passes locally (Node 24.13.1, pnpm 11.20.0, Docker 29.2.1): lint, typecheck, clean build, both isolated databases recreated, all 25 critical API suites (each in its own process, skips forbidden), remaining package tests, removed-component check, guards, audit (2 high, both pre-existing and ignored by repo config).
- [x] Behaviour verified beyond the test suite: the new integration suite drives the real `TurnDispatcher`, `StoryWorkspaceService`, `TurnGitEvidenceService`, and `StoryCollisionService` against Postgres with the fake workspace runtime, with an engine that writes exactly the files each turn should be blamed for, so the evidence rows the guard reads are produced by the same capture path as production.
- [x] Documentation updated: `reference/lifecycle.md` (turn evidence section), `guides/operate-story.md` (what the timeline entry means), and the documentation contract test now requires `story.collision_detected` to be documented.

Commands run:

```bash
DATABASE_URL=postgres://facility:facility@127.0.0.1:5461/facility_test \
  pnpm --filter @facility/api exec vitest run test/story-collisions.test.ts               # 2 passed
DATABASE_URL=postgres://facility:facility@127.0.0.1:5461/facility_test \
  pnpm --filter @facility/api exec vitest run test/story-collisions.integration.test.ts   # 2 passed
pnpm verify                                                                               # exit 0
```

Coverage added:

- `services/api/test/story-collisions.test.ts`: sorted intersection; empty block without collisions; path cap with `(+N more)`; story cap with the "N more stories" line.
- `services/api/test/story-collisions.integration.test.ts`:
  - two stories change `src/shared.ts`; the first turn of each is unwarned (no evidence yet); the next turn of the second story gets the prompt block naming the first story and only the shared path, one `story.collision_detected` fact with the expected external key and data, one `turn.collisions_detected` event, no attention item, and the story is not in `attention`;
  - the overlap is symmetric: the first story's next turn is warned about the second;
  - three stories share `lib/core.ts`; after one is archived and another's branch is recorded as a merged pull request, the remaining story's next turn has no block and no facts.

Not run: the Docker-backed workspace E2E tier. This change does not touch workspace execution boundaries.

## Open questions

- Should merged-branch exclusion also honour `github_branches.deleted_at`, for branches deleted without a mirrored pull request? Left out to keep the rule explainable in one sentence; it is a one-clause change.
- Would you rather the prompt block ask the agent to add an explicit note to the PR body, instead of "when it matters"? I kept the softer wording to avoid noisy PR descriptions for trivial overlaps.

### DIFF
--- a/apps/docs/docs/guides/operate-story.md
+++ b/apps/docs/docs/guides/operate-story.md
@@ -44,7 +44,10 @@ session, workspace, branch, and initial SHA started each turn; the final SHA, co
 dirty state; artifacts and attention; and the mirrored branch, pull request, reviews, and checks.
 An entry linked to a `turn_id` has an exact Facility attribution. An `external` GitHub actor means
 Facility associated the event with the story but could not prove that a particular turn produced
-it.
+it. A `story.collision_detected` entry means another open story in the project has changed the
+same files; the agent was told which paths overlap before its turn started, and the two pull
+requests are likely to conflict. Nothing is blocked, but it is worth deciding early which one
+lands first.
 
 When the story needs attention:
 
--- a/apps/docs/docs/reference/lifecycle.md
+++ b/apps/docs/docs/reference/lifecycle.md
@@ -54,6 +54,15 @@ reconciliation. A matching final SHA links the fact to an exact turn. Facts that
 only through the story's branch, pull request, or closing issue remain story-level external changes
 rather than being attributed to an agent.
 
+Before the engine starts, Facility also compares the changed files recorded for this story with
+the changed files of every other active story in the project that has an open branch. Stories that
+are done, archived, deleted, or whose branch belongs to a merged pull request are ignored. When two
+stories touched the same paths, a `story.collision_detected` fact is recorded on the timeline for
+each overlapping story, with up to 20 overlapping paths and the total count, and the prompt tells
+the agent which paths other open branches are changing. The check is advisory: it never blocks or
+fails a turn, does not create attention, and a story with no completed evidence yet cannot collide
+because its files are not known. A detection failure is recorded as `turn.collision_check_failed`.
+
 The story response orders these facts with turn activity, artifacts, attention, and creation in a
 single timeline. Source time and observation time are both retained because an event discovered by
 reconciliation may be older than the time Facility first saw it.
--- a/apps/docs/test/documentation-contract.test.mjs
+++ b/apps/docs/test/documentation-contract.test.mjs
@@ -146,6 +146,7 @@ test("project capabilities and delivery evidence are inspectable", () => {
     "commits",
     "changed files",
     "periodic reconciliation",
+    "story.collision_detected",
     "single timeline",
   ]) {
     assert.match(
--- a/apps/web/app/(app)/projects/[projectId]/stories/[number]/page.tsx
+++ b/apps/web/app/(app)/projects/[projectId]/stories/[number]/page.tsx
@@ -365,6 +365,10 @@ function timelineSummary(type: string, data: Record<string, unknown>) {
   if (type === "github.check_observed") {
     return `${text("name") ?? "check"} · ${text("conclusion") ?? text("status") ?? "unknown"}`;
   }
+  if (type === "story.collision_detected") {
+    const extra = Number(data.overlapCount ?? 0) - count("overlappingPaths");
+    return `overlaps "${text("title") ?? "another story"}" on ${String(data.overlapCount ?? "?")} files · ${text("branch") ?? "unknown branch"}${extra > 0 ? ` · ${extra} not listed` : ""}`;
+  }
   if (type === "artifact.recorded") return text("label") ?? "Artifact recorded.";
   if (type.startsWith("attention.")) return text("title") ?? type;
   if (type === "turn.failed") return text("error") ?? "Turn failed.";
--- a/services/api/src/turns/collisions.ts
+++ b/services/api/src/turns/collisions.ts
@@ -1,0 +1,172 @@
+import { type FacilityDb, githubPullRequests, stories, turnGitEvidence } from "@facility/db";
+import { and, eq, inArray, isNotNull, isNull, ne, sql } from "drizzle-orm";
+
+/** Paths listed per colliding story in the prompt and the evidence event. */
+export const MAX_COLLISION_PATHS = 20;
+/** Colliding stories reported per turn; the rest are counted. */
+export const MAX_COLLISION_STORIES = 10;
+
+const ACTIVE_STORY_STATUSES = ["ready", "working", "attention", "review"] as const;
+
+export type StoryCollision = {
+  storyId: string;
+  title: string;
+  provider: string;
+  externalId: string;
+  branch: string;
+  status: string;
+  overlappingPaths: string[];
+  overlapCount: number;
+};
+
+/**
+ * Detects other active stories in the same project whose recorded Git evidence changed files this
+ * story also changed. It is advisory only: it reads evidence that already exists and never blocks
+ * dispatch. A story with no completed evidence yet cannot collide, because its files are unknown.
+ */
+export class StoryCollisionService {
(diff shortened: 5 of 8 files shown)


Lob26:
The posture here is right and worth saying out loud: advisory by construction, never blocks, a failed check records `turn.collision_check_failed` and the turn continues. The external key `turn:<turnId>:collision:<otherStoryId>` makes a retried turn idempotent, the merged-branch filter goes to the mirror instead of trusting story status, and `jsonb_path_query_array` keeps the `changedFiles` blobs in Postgres rather than hauling them into Node. That last one is the detail most implementations miss.

One concern, on the hot path. `changedPaths` bounds its *output* but not its *input*:

```ts
.where(and(..., inArray(turnGitEvidence.storyId, storyIds), isNotNull(completedAt), isNull(captureError)))
```

That reads every completed evidence row for every active story in the project, on every dispatch, before the engine starts. `MAX_COLLISION_PATHS` and `MAX_COLLISION_STORIES` cap what is reported; nothing caps what is read or what is materialized into the `Set`s. A long-lived project accumulates one row per turn per story and nothing prunes them, so the cost grows with project age rather than with the size of the collision.

The good news is that the bound is nearly free: `turn_git_evidence_story_completed_idx` is already `(org_id, story_id, completed_at desc)`, so taking only the most recent N evidence rows per story is an `ORDER BY` plus a `LIMIT` on an index that already sorts that way. No migration.

Smaller note: `mine` unions every turn's evidence for the story, so a file touched in an early turn and later reverted still counts as an overlap. Probably fine for something advisory, but it may be worth a line in the guide so the reader knows the set is cumulative rather than current.

If a patch for the bound would help, happy to put one up against your branch. Otherwise it is yours and I will stay out of the way.

PR #323, opened 2026-09-07
fix(previews): re-check workspace execution on every preview request

Lob26:
## The gap

A preview session lives for an hour and proxies **every HTTP method plus WebSocket traffic** into the running workspace. Authorization for it was decided once — by `config: { permission: "workspaces:execute" }` on the open route — and never asked again.

`WorkspacePreviewService.assertMembership` already selected the permissions and then dropped them on the floor:

```ts
const member = (
  await this.db
    .select({ permissions: roles.permissions })   // fetched…
    ...
)[0];
if (!member) throw invalidAccess();               // …and never read
```

So a member whose role is downgraded — or swapped for a read-only one — keeps driving the workspace until the session expires on its own. The membership row and the user row both stay valid in that state, which is exactly what the surrounding checks tolerate: `users.status = 'disabled'` is caught, a revoked session is caught, an expired token is caught. A revoked *permission* was not.

Two documented claims cover this and were not true:

- `apps/docs/docs/self-host/authentication.md`: "each proxied request revalidates the user, project membership, workspace, service, expiry, and revocation status."
- `apps/docs/docs/reference/security.md`: "HTTP and WebSocket requests are authorized continuously rather than only when a preview is opened."

`AGENTS.md` also names `revoked` as a denial path that authorization changes must cover.

## The change

Re-read the permission on `exchange` and `authorize`, and rename the helper to say what it now does. `workspaces:execute` is used because that is what the open route requires — this keeps the session's authority equal to the authority that created it, rather than widening or narrowing it.

The open route is deliberately left alone: it is already guarded, and duplicating the check there would add a query without adding safety.

## Test

`services/api/test/workspace-preview.integration.test.ts` gains one case that opens and consumes a session, then downgrades the role to `["previews:read"]` — leaving the user active and the membership row intact — and asserts the proxy path closes. It also asserts a fresh handoff cannot be exchanged while the permission is gone, and that restoring the permission restores access, so the test fails for the permission and not for a side effect.

Without the change, the test fails on the first assertion: `authorize` returns the live session.

## Verification

Run on Windows 11, Node 26, against the compose Postgres.

- `pnpm lint` — clean
- `pnpm guards` — 2 guards, 0 failed
- `pnpm check:unused` — clean
- `vitest run test/workspace-preview.integration.test.ts test/workspace-preview.test.ts` — 13 passed
- Full `services/api` suite: 258 passed, 14 failed — every failure also fails on unmodified `main` in this environment (the POSIX-shell fakes in `workspace-vercel-bootstrap`, `agent-engines`, `facility-012.e2e`, `project-environment`, `turn-dispatcher`, i.e. the `#227`/`#241` Windows family), plus one load-dependent flake in `workspace-runtime.test.ts` that passes 3/3 in isolation and only touches `FakeWorkspaceRuntime`. I did not get a clean-baseline Linux run, so those are worth a second look on CI rather than taking my word for it.

> Claude Code helped

### DIFF
--- a/services/api/src/workspaces/preview.ts
+++ b/services/api/src/workspaces/preview.ts
@@ -1,5 +1,5 @@
 import { createHash, randomBytes } from "node:crypto";
-import { newId } from "@facility/core";
+import { can, newId } from "@facility/core";
 import {
   type FacilityDb,
   orgMembers,
@@ -17,6 +17,9 @@ import type { ProjectEnvironmentService, ProjectManifestSource } from "./project
 import type { WorkspaceLocator, WorkspaceRuntime } from "./runtime.js";
 
 const SESSION_TTL_MS = 60 * 60 * 1_000;
+
+/** The permission the open route requires; a live session must keep holding it. */
+const PREVIEW_PERMISSION = "workspaces:execute";
 
 export class WorkspacePreviewError extends Error {
   constructor(
@@ -123,7 +126,7 @@ export class WorkspacePreviewService {
         .returning()
     )[0];
     if (!consumed) throw invalidAccess();
-    await this.assertMembership(consumed.orgId, consumed.userId);
+    await this.assertPreviewAccess(consumed.orgId, consumed.userId);
     return consumed;
   }
 
@@ -144,7 +147,7 @@ export class WorkspacePreviewService {
         .limit(1)
     )[0];
     if (!session) throw invalidAccess();
-    await this.assertMembership(session.orgId, session.userId);
+    await this.assertPreviewAccess(session.orgId, session.userId);
     return session;
   }
 
@@ -218,7 +221,11 @@ export class WorkspacePreviewService {
     return { story, workspace };
   }
 
-  private async assertMembership(orgId: string, userId: string) {
+  /**
+   * A preview session outlives the request that opened it, so the permission
+   * that opened it is re-read on every use: membership alone is not access.
+   */
+  private async assertPreviewAccess(orgId: string, userId: string) {
     const member = (
       await this.db
         .select({ permissions: roles.permissions })
@@ -234,7 +241,7 @@ export class WorkspacePreviewService {
         )
         .limit(1)
     )[0];
-    if (!member) throw invalidAccess();
+    if (!member || !can(member.permissions, PREVIEW_PERMISSION)) throw invalidAccess();
   }
 }
 
--- a/services/api/test/workspace-preview.integration.test.ts
+++ b/services/api/test/workspace-preview.integration.test.ts
@@ -212,6 +212,37 @@ describe("workspace preview session security", async () => {
     });
   });
 
+  it("stops proxying as soon as the member loses workspace execution", async () => {
+    const opened = await service.open({ orgId, projectId, storyId, userId, service: "web" });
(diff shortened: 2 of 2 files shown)


javiertoledo:
Reviewed this and pushed one commit to the branch (`maintainer_can_modify`): `8e0af84`, unit coverage only. `preview.ts` is untouched — the fix is right and I could not find a reason to change it.

**Verified against `main` rather than taken from the description.** `assertMembership` really did `select({ permissions: roles.permissions })`, join `roles`, and then only check `if (!member)`. The permissions came back from the database on every request and were discarded.

I also checked the things that would make the fix wrong rather than just confirming it works:

- All three traffic paths reach `assertPreviewAccess` — `exchange`, the HTTP proxy, and the WebSocket upgrade all route through `authorize()`. No path proxies on a session that skipped the check.
- `can()` honours `*`, exact grants and `resource:*`, and it is the same helper `app.ts:279` uses for routes. So this cannot deny a role the route layer would admit.
- The `workspaces:read` GET routes are the environment-variable ones, not previews. Preview open is `workspaces:execute` only, so re-checking that exact permission is consistent and there is no availability regression for read-only members.

**What I added, and why it isn't redundant with your integration test.** Yours is good — it revokes the permission while keeping the membership row and user active, which is what proves it is the permission being read and not membership, and it restores the grant afterwards to show it isn't a permanent lockout. CLAUDE.md asks for unit *and* deterministic integration tests on an RBAC change though, and the unit half is the half that always runs: `workspace-preview.integration.test.ts` skips itself when Postgres is unreachable, so on a machine without a database nothing exercised the new check.

`workspace-preview.test.ts` already builds the service against a faked `FacilityDb`, so `8e0af84` follows that pattern rather than inventing one — five cases driving `authorize()` directly: the exact grant, the `workspaces:*` and `*` wildcards, the revoked role, and no active membership. Only the revoked case fails against the previous check; the other four passed before this branch and must keep passing, or the fix would be denying people it should admit.

Unit suite 13 passed, `tsc --noEmit`, `biome check` and `node guards/run.mjs` clean. I did not run the integration file locally — Postgres is unreliable on this machine — so CI is the authority on that half.

**One thing I deliberately did not fix here**, filed as #383: a WebSocket is authorized once at upgrade and then proxies frames with no further checks, so a demoted member keeps a live bidirectional channel until something else closes it. Same class this PR fixes, and not made worse by it. Closing it needs a timer or a frame hook plus decisions about interval and close behaviour — that would turn 43 reviewable lines into a design review. Enforcing `SESSION_TTL_MS` on an open socket has the same shape and probably the same answer.

One smaller thing I looked at and left alone: `exchange` consumes the one-time token before checking access, so a revoked member burns their link. Fail-closed is the right order and `open` mints another, so it costs nothing worth a commit.

Lob26:
Thank you, and the gap you closed is the one I should have seen: my integration test skips itself when Postgres is unreachable, so on a machine without a database nothing exercised the new check at all. Unit coverage is the half that always runs. I have taken that as the rule for the rest of my open PRs rather than just this one. Happy to pick up #383 as its own change if it is free, with the `SESSION_TTL_MS` question folded in, since both come down to the same thing: a socket authorized once and never asked again.

Issue #280, opened 2026-09-02
[0.12] Run project environments and authenticated previews from workspaces


adrian-lorenzo:
## Goal

Run the full development environment inside the story workspace and let agents and people test the same running services.

## Project contract

Add a small repository-owned contract for:

- repositories and the default repository
- setup command
- start and stop commands
- readiness check
- declared services and ports
- optional seed command
- browser test command
- names of project secrets and environment variables required by setup and services

Support Docker or Docker Compose through the runtime chosen by the architecture decision. A project may use ordinary commands when containers are not needed.

Store a checksum of the contract and relevant setup inputs. Setup can be reused when the checksum is unchanged. A maintainer can request a clean setup without deleting the Git worktree or engine sessions.

The control service injects declared project secrets and environment variables when setup or services run. They are available to every agent in the project and must not be written to the durable volume, story output, command logs, or artifacts.

## Service behavior

Facility reports whether setup, start, readiness, browser testing, and preview routing succeeded. Failures should name the command, service, exit status, and recent output.

The agent can run a headless browser against internal service addresses. Screenshots, traces, and reports can be attached to the story as artifacts.

## Preview behavior

Expose only declared service ports. Opening a preview wakes compute, starts services if needed, waits for readiness, and returns an authenticated URL. Active preview sessions keep compute awake.

Preview access is limited to authenticated project maintainers. It expires and can be revoked. Routing must isolate project, story, workspace, and port. It supports WebSockets when the service declares them.

The preview points to the live workspace. Do not build a separate image or create a second preview sandbox.

## Acceptance criteria

- [ ] A repository contract defines setup, services, readiness, ports, seeds, and browser tests.
- [ ] Invalid contracts fail before any partial setup command runs.
- [ ] Setup reuse is controlled by a recorded checksum.
- [ ] The reference fixture starts its application, API, and database through Compose.
- [ ] Seed data is available after compute replacement.
- [ ] A headless browser completes the fixture's user flow inside the workspace.
- [ ] Browser screenshots and traces are stored as story artifacts.
- [ ] An authorized user can open the live service from another computer.
- [ ] Preview expiry and revocation work without deleting the workspace.
- [ ] Cross-tenant, wrong-port, malformed, expired, and replayed preview requests are denied by deterministic integration tests.
- [ ] WebSockets stay connected through the proxy, and reloading the browser returns the same workspace service.
- [ ] Injected secrets are absent from the durable volume, story output, command logs, and artifacts.
[...]

PR #328, opened 2026-09-07
fix(github): name the kickstart failure instead of returning a bare 500

### LINKED ISSUE #211: Kickstart returns "Internal server error" when processing a repository
### What happened

Hi! I was able to complete the GitHub authentication successfully and reach the Kickstart flow.

The issue appears when trying to use the repository:

panteraimperial-hub/facility-test

In Kickstart a project → pick a repository → preview the assets → open the PR , Facility returns:

Internal server error

So, at least from my testing, GitHub login/authentication is working, but the Kickstart flow fails when Facility tries to process the repository.

### How to reproduce

1. Run Facility locally.
2. Sign in successfully using GitHub authentication.
3. Open the "Kickstart a project" flow.
4. Enter/select the repository: panteraimperial-hub/facility-test
5. Attempt to continue to preview the assets / create the project.

Expected:
Facility should load and process the repository and continue to the asset preview.

Actual:
Facility displays: "Internal server error".

GitHub authentication itself is working correctly; the failure occurs when Kickstart attempts to process the repository.

### How are you running Facility?

Local development (pnpm dev)

### Version

ca1c9df

### Evidence

The Kickstart page displays:

Internal server error

Repository:
panteraimperial-hub/facility-test

Authentication with GitHub completes successfully and Facility recognizes the signed-in GitHub account. The error occurs only when attempting to process the repository through Kickstart.

Lob26:
Closes #211.

## The report, and what actually happens

#211 reaches kickstart with working GitHub authentication, selects a repository, and gets `Internal server error`. There is nothing else to go on — not whether the repository, the installation, or Facility is at fault. That is the first flow a new installation runs, so it is the worst possible place for an opaque answer.

There are two causes behind the one symptom.

**1. Octokit reports `status`; the error handler reads `statusCode`.**

```ts
// services/api/src/app.ts
const status = typeof err.statusCode === "number" ? err.statusCode : 500;
if (status >= 500) { /* logged, masked as "Internal server error" */ }
```

`@octokit/request-error` sets `status`, not `statusCode`. So every refusal GitHub can give — 404, 403, 409, 429 — falls into the 500 branch and is masked. The codebase already knows this in one place; `kickstartRepo` reads `(error as { status?: number }).status` when handling a 422 from `createBranch`, and nowhere else.

For the reported case the failing call is `getDefaultBranchSha()` → `repos.getBranch`, which answers 404 on a repository with no commits. `facility-test` reads like a freshly created repository, which fits: `readRepoFiles` swallows every `getContent` failure, so preview renders happily, and the first call that actually needs a base commit is the one that dies.

**2. Kickstart's own refusals were bare `Error`s**, which have no `statusCode` at all — including `"Repository already contains the Facility 0.12 kickstart files"`. That is a conflict the caller can act on, delivered as a server fault.

## The change

Map both, in the one boundary that owns the GitHub call.

| Condition | Before | After |
|---|---|---|
| Base ref unreadable (no commits, missing branch, repo no longer in the installation) | 500 | 404 `kickstart_repository_unreachable` |
| GitHub reports the repository empty | 500 | 409 `kickstart_repository_empty` |
| Installation refuses the request | 500 | 403 `kickstart_repository_forbidden` |
| Installation rate limited | 500 | 429 `kickstart_github_rate_limited` |
| Repository has no installation | 500 | 409 `github_installation_missing` |
| Installation suspended or foreign | 500 | 409 `github_installation_unavailable` |
| Kickstart files already present | 500 | 409 `kickstart_already_applied` |

Each message names the repository, the ref, and the remedy. **Statuses with no advice attached are passed through untouched**, so a genuine upstream fault is still logged and masked by `sendError` — the fix narrows what gets a 500, it does not stop masking.

## Test

New `services/api/test/kickstart-failures.test.ts`, six cases against a fake Octokit that throws `status`-carrying errors the way `@octokit/request-error` does — no daemon, no network, runs in the default suite. Five fail on unmodified `main`; the sixth is the regression guard that an unmapped status *stays* masked, and it passes both ways by design.

## Found but not fixed

`readRepoFiles` swallows every `getContent` error, so **kickstart preview succeeds on a repository that apply cannot use**. #211's reporter gets all the way through "preview the assets" before the failure appears. Making preview resolve the base ref would surface the same diagnosable error one step earlier, which reads like the right behaviour and matches #280's "invalid contracts fail before any partial setup command runs". I left it out to keep this to one change; happy to add it here or open it separately, whichever you prefer.

Distinguishing "empty repository" from "branch does not exist" would need one extra `repos.get` call on the error path. I did not add it — the 404 message names both possibilities rather than inventing a diagnosis.

## Verification

Windows 11, Node 26, compose Postgres.

- `pnpm lint` — clean
- `vitest run test/kickstart-failures.test.ts test/github-client.test.ts` — 8 passed
- Full `services/api` suite: 261 passed, 13 failed — the same 13 that fail on unmodified `main` in this environment (the POSIX-shell fakes in `workspace-vercel-bootstrap`, `agent-engines`, `facility-012.e2e`, `project-environment`, `turn-dispatcher`, i.e. the #227/#241 Windows family). Verified by stashing the change and re-running.

I could not reproduce against a live GitHub App, so the status-to-condition mapping is from the REST documentation and Octokit's error shape rather than from observed responses. The fake reproduces the shape; a real empty-repository run would confirm the 404 path end to end.

> Claude Code helped

### DIFF
--- a/services/api/src/github/kickstart.ts
+++ b/services/api/src/github/kickstart.ts
@@ -1,6 +1,7 @@
 import { renderWorkspaceKickstart, sha256Hex } from "@facility/core";
 import { type FacilityDb, githubInstallations } from "@facility/db";
 import { eq } from "drizzle-orm";
+import { ApiError } from "../errors.js";
 import type { AppConfig, Principal } from "../types.js";
 import { FacilityGithubClient, type GithubClientFactory, type TreeItem } from "./client.js";
 import { readRepoFiles } from "./repo-files.js";
@@ -35,7 +36,13 @@ export async function createGithubClientForRepo(
   factory: GithubClientFactory,
   repository: GithubRepositoryRow,
 ): Promise<FacilityGithubClient> {
-  if (!repository.installationId) throw new Error("Repository has no GitHub installation");
+  if (!repository.installationId) {
+    throw new ApiError(
+      409,
+      "github_installation_missing",
+      `${slug(repository)} is not connected to a GitHub App installation. Reconnect the repository before running kickstart.`,
+    );
+  }
   const installation = (
     await db
       .select()
@@ -44,7 +51,11 @@ export async function createGithubClientForRepo(
       .limit(1)
   )[0];
   if (!installation || installation.orgId !== repository.orgId || installation.suspendedAt) {
-    throw new Error("GitHub installation is unavailable");
+    throw new ApiError(
+      409,
+      "github_installation_unavailable",
+      `The GitHub App installation for ${slug(repository)} is suspended or belongs to another organization.`,
+    );
   }
   return new FacilityGithubClient(await factory(installation.installationId), {
     owner: repository.owner,
@@ -59,9 +70,12 @@ export async function kickstartPreview(
   repository: GithubRepositoryRow,
   answers: KickstartAnswers,
 ) {
+  const ref = answers.defaultBranch ?? repository.defaultBranch;
   const client = await createGithubClientForRepo(db, factory, repository);
-  const existing = await readRepoFiles(client, repository.defaultBranch);
-  const detection = detectWorkspace(existing, answers.defaultBranch ?? repository.defaultBranch);
+  const existing = await readRepoFiles(client, repository.defaultBranch).catch((error) => {
+    throw kickstartFailure(error, repository, ref);
+  });
+  const detection = detectWorkspace(existing, ref);
   const workspaceAnswers = workspaceKickstartAnswers(
     repository,
     answers,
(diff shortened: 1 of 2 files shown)


adrian-lorenzo (changes requested):
Thanks for tackling this.

Please distinguish [403 rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api#exceeding-the-rate-limit) from permission failures using the existing helper. Also avoid treating every 409 as an empty repository—a later ref conflict gets that message too. Add regression coverage for both.

Lob26:
Addressed at eee45b1. Both were one status carrying two conditions, and my message named the wrong remedy half the time. Kickstart now asks `githubRateLimitRetryAt` first, so a throttled 403 reports the reset instead of sending someone to the App permissions. For 409 I split `applyKickstart` into resolving the base and writing the refs: a 409 while resolving is an empty repository, a 409 once the branch is being written is another writer having moved it. Three regressions, all red without the change. Thanks for the pointer to the helper.

Issue #19, opened 2026-07-20
Simplify deployment into an easily installable cloud-agnostic bundle


javiertoledo:
## Problem

The self-host path today is a multi-hour, AWS-specific Terraform runbook (ECS, RDS, CloudFront, Secrets Manager, one-shot bootstrap tasks). That is a hard blocker for adoption: someone evaluating the platform should reach a working governed run in minutes, not in an afternoon of infrastructure work.

## Scope

- Define the minimal deployable unit (api + worker + gateway + web + Postgres) and ship it as a **cloud-agnostic bundle**: one `docker compose up` (or equivalent) driven by a single configuration file.
- Automate what is manual today: migrations, master key, operator bootstrap, and GitHub App creation (the App manifest flow can register the App, its permissions and its webhook programmatically — see #34).
- Keep the Terraform path as the "production reference" profile; the bundle is the adoption path.
- Keep tenancy assumptions out of the bundle so it can also serve as a substrate for hosted deployments.

## Acceptance

- [ ] Someone with Docker and a GitHub organization gets a working governed run on a test repository in under 15 minutes, following one documentation page.
- [ ] No step requires reading Terraform.

Lob26:
Two of the acceptance criteria here have moved.

#333 fixes why `docker compose up` does not start the API today. The api image runs in production mode, the compose file defaults the preview origin to plain HTTP, and production validation refuses that. It also adds the single documentation page this issue asks for.

#336 makes the operator bootstrap runnable from the bundle, so binding the first owner no longer needs a local Node toolchain or ten flags typed correctly.

What is still open is creating the GitHub App by hand. Nothing in the code uses the App manifest flow yet, and that is the step that costs the most minutes. Happy to take it if nobody else is on it.

PR #333, opened 2026-09-07
fix(config): boot the single-host bundle on loopback origins

Lob26:
## What changes

`docker compose up` starts the bundle. Today it does not: the API exits during
config validation before it ever listens.

The `api` container runs the `api` image, which sets `NODE_ENV=production`
(`Dockerfile:86`), and `docker-compose.yml` defaults `FACILITY_PREVIEW_URL` to
`http://preview.localhost:4400`. Production validation refuses any preview
origin that is not HTTPS:

```
FACILITY_PREVIEW_URL must use HTTPS in production
```

So the command in the Compose file's own header comment —
`SECRET_MASTER_KEY=$(openssl rand -base64 32) docker compose up -d` — cannot
bring up an instance.

An instance whose every origin is loopback HTTP is now exempt from that
requirement. This is the carve-out the interactive OAuth block in the same
`superRefine` already makes for its own URL set, applied to the preview origin
that did not have it. `isLoopbackHostname` additionally accepts names under
`.localhost`, which RFC 6761 §6.3 reserves for the loopback interface —
Facility needs more than the bare name because the preview origin must stay a
registered site of its own.

A second commit adds `self-host/bundle.md`, the page that describes the path.

## Why

Requiring HTTPS on loopback refuses a configuration that has no transport to
protect: there is no name to obtain a certificate for and nothing is reachable
off the machine. Any origin that leaves the machine puts the whole set back
under the requirement, and the separate-registered-site rule is untouched —
previews stay isolated from the control plane on loopback too.

**Blast radius, stated rather than buried.** `isLoopbackHostname` has three call
sites. Two widen: `FACILITY_INSECURE_DEV` may now be enabled on a `.localhost`
control origin, and `MCP_PUBLIC_URL` may be HTTP there. Both already accepted
`localhost` itself, neither leaves the host, and `FACILITY_INSECURE_DEV` remains
refused under `NODE_ENV=production` regardless.

`docker-compose.yml` has carried the whole 0.12 control plane since #289, but no
page describes it: the quickstart is the from-source path and `production.md`
mentions the file in one clause. The new page is the single documentation page
#19 asks for, and it states the limits rather than leaving them to be
discovered — GitHub cannot deliver webhooks to `localhost`, the bundle is
TLS-free only while every origin stays loopback, and Facility never reclaims
workspace volumes by age.

Related: **#183**. Its root cause — a shipped template value the platform's own
validation rejects — was fixed in #159 by the `OptionalNonEmpty` preprocess.
This is the same failure class one layer up, and the new test pins the `.env.example`
half so it cannot come back. #183 looks closeable.

## Verification

The regression fails without the fix, for the right reason:

```
× boots the single-host bundle on loopback origins
AssertionError: [ { "path": ["FACILITY_PREVIEW_URL"],
  "message": "FACILITY_PREVIEW_URL must use HTTPS in production" } ]
```

Coverage pins the shipped defaults rather than a paraphrase of them: the Compose
environment boots; a published control plane with a loopback preview does not; a
loopback control plane with a published preview does not; the bare control
hostname is still refused as a preview origin; and `.env.example` is parsed as
dotenv delivers it and asserted to boot with only the master key filled in.

```
vitest run test/config.test.ts                          20 passed
vitest run config + workspace-preview + identity-provider + oauth
                                                        72 passed | 1 skipped
tsc --noEmit (@facility/api)                            clean
```

identical 92/24 on a clean `main`. The failures are Windows environment noise —
the patched `image-size` cases, `tar` failing to resolve `C:`, and the registry
publication tests. Beyond the suite, the defect and the fix were both confirmed
against the real file: `docker compose --env-file .env.example config` reproduces
the interpolation path, and the failing validation was reproduced with the exact
environment the Compose file hands the container.

> 🤖 Claude Code helped

### DIFF
--- a/apps/docs/docs/self-host/bundle.md
+++ b/apps/docs/docs/self-host/bundle.md
@@ -1,0 +1,129 @@
+---
+title: Compose bundle
+---
+
+# Compose bundle
+
+The bundle is the adoption path: one `docker compose up` brings up the whole control plane — API
+with embedded MCP and webhooks, worker, web UI, PostgreSQL, and the workspace runner image — from
+the repository's `docker-compose.yml`. It needs no cloud account and no Terraform.
+
+Use the [quickstart](quickstart.md) instead when working on the Facility source, and the
+[AWS reference deployment](aws.md) when a hosted control plane is the goal. The bundle runs every
+service on one host with one Docker daemon, so it is an evaluation and small-team shape, not a
+resilient deployment.
+
+## Prerequisites
+
+Docker with Compose v2, a running daemon the current user can reach, and a GitHub organization
+whose repositories Facility may automate. Story workspaces hold repository checkouts, dependencies,
+nested images, and persistent volumes; keep several gigabytes of disk free.
+
+## Start the control plane
+
+The master key encrypts every stored credential. Generate it once and keep it: an instance that
+loses its key cannot decrypt the project secrets it already holds.
+
+```bash
+git clone https://github.com/theam/facility.git
+cd facility
+printf 'SECRET_MASTER_KEY=%s\n' "$(openssl rand -base64 32)" >> .env
+docker compose up -d
+```
+
+The first run builds the API, web, and runner images. `migrate` applies the schema and must exit
+zero before the API and worker start; `runner-image` builds the workspace image the worker later
+hands to stories. Both are one-shot services, so `docker compose ps` showing them as exited is the
+expected steady state.
+
+Check the control plane:
+
+```bash
+curl --fail http://localhost:4400/health
+curl --fail http://localhost:4400/readyz
+```
+
+| Surface | URL |
+| --- | --- |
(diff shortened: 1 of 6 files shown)


adrian-lorenzo (changes requested):
Thanks.

Please bind the Compose API and web ports to loopback before allowing HTTP. Localhost URLs do not restrict network access. Add a regression check for those bindings.

Lob26:
Addressed at aa4bf30. You are right, and worse than you put it: my description claimed nothing was reachable off the machine, while `4400:4400` published plaintext sessions and preview traffic to the whole network segment. Both ports are now bound to `127.0.0.1`; container traffic uses the Compose network, so nothing internal was published anyway. The regression lives in `config.test.ts` beside the carve-out, because the premise and the rule sitting in different files is how the premise rotted unnoticed. Thank you for not letting that through.

PR #334, opened 2026-09-07
fix(agents): run a schedule that fell behind once, not once per missed occurrence

### LINKED ISSUE #330: A scheduled agent replays every occurrence it missed while the worker was down, one paid run per tick
## What happens

`AgentScheduler.claim` advances a schedule by exactly one cron step, computed from the schedule's own stored `nextRunAt` rather than from now:

```ts
// services/api/src/agents/scheduler.ts
private async claim(schedule: typeof agentSchedules.$inferSelect, now: Date) {
  const nextRunAt = nextOccurrence(schedule.cron, schedule.timezone, schedule.nextRunAt);
  ...
}
```

`tick()` selects every schedule with `nextRunAt <= now` and claims one occurrence each, and the worker runs `agent.schedules` on `* * * * *`:

```ts
await boss.schedule("agent.schedules", "* * * * *", {});
```

So a schedule that fell behind catches up one occurrence per minute, and each caught-up occurrence starts a real story turn.

## Why it matters

The backlog is paid work, not bookkeeping. `stories.start` provisions a workspace and dispatches an engine, so every replayed occurrence is a provider call charged to the project's monthly budget.

An hourly schedule after a day of worker downtime fires **24 turns, one per minute** — a burst against a budget sized for one run an hour, in a window narrow enough that the preflight check sees the earlier ones as still in flight. A five-minute schedule after the same outage is 288. Nothing bounds it: the loop replays as many stale occurrences as the outage contains.

Downtime is not exotic here. `pnpm dev` restarts the worker whenever `tsx watch` sees a file change, and #35 describes the same class of interruption on a deployed instance.

The occurrences being replayed are also, by definition, stale. A `security-audit` at 05:00 Monday that runs at 14:32 Tuesday is not the audit anyone asked for; it is the same audit against a different repository state, and the run it would have preceded already happened.

## Reproduction

No live provider needed — `nextOccurrence` is exported and pure:

```js
import { nextOccurrence } from "./services/api/src/agents/scheduler.js";

// A schedule last due at 06:00, a worker that returns 24 hours later.
let nextRunAt = new Date("2026-09-06T06:00:00Z");
const now = new Date("2026-09-07T06:00:00Z");
let fired = 0;
while (nextRunAt <= now) {
  nextRunAt = nextOccurrence("0 * * * *", "UTC", nextRunAt); // what claim() does
  fired += 1;
}
console.log(fired); // 24 — one dispatch per tick until the backlog drains
```

## What is already right

Worth saying, because the fix should not disturb it: the claim is a proper compare-and-swap on `(nextRunAt, lastScheduledAt)`, so two workers cannot both claim the same occurrence, and `syncProject` takes a `pg_advisory_xact_lock` before rewriting a project's schedules. The concurrency story is sound. The gap is only in *which* occurrence is claimed after a gap.

## Possible shapes

I have no strong preference between these and would rather agree on the semantics than send a patch that picks one silently.

**Coalesce.** Compute the next occurrence from `now` instead of from the stored `nextRunAt`, keeping the CAS on the observed value. A backlog of any size collapses to one run. Three lines, and it makes the loop level-triggered: it converges on "this schedule is due" instead of replaying every edge it missed.

**Bound it.** Keep catch-up but skip occurrences older than a deadline, the way a Kubernetes CronJob uses `startingDeadlineSeconds` and treats more than 100 missed schedules as an error worth surfacing rather than absorbing.

**Make it the manifest's choice.** A trigger declares whether a missed occurrence is worth running late. More faithful, and more surface than the problem probably deserves today.

Coalescing is what I would pick: for every agent kickstart ships — `security-audit`, `ci-doctor`, `pr-reviewer` — a missed run wants to happen once, now, not N times against a repository that has moved on. But "run every missed occurrence" is a legitimate contract for some schedules, and if that is the intent then the bug is that nothing bounds or reports the backlog.

Either way the loop should be able to say it fell behind. `tick()` returns `{ projects, due, scheduled, failures }`; `due` counts rows, not how stale they were, so a 24-occurrence catch-up and a normal minute are indistinguishable in the log.

## Scope

`services/api/src/agents/scheduler.ts` only. The file is untouched by the 0.12 rebuild and by every open pull request, and `services/api/test/agent-scheduler.test.ts` already exercises `nextOccurrence` with a fixed clock, so whichever semantics you choose is deterministically testable without new harness.

Happy to implement it once the intended behaviour is settled.

> Claude Code helped

Lob26:
## What changes

A schedule that fell behind runs **once** when the worker comes back, instead of
once per missed occurrence.

`AgentScheduler.claim` advanced a due schedule by exactly one cron step from its
own stored `nextRunAt`. The row therefore stayed due after the claim, and `tick`
— which the worker runs on `* * * * *` — dispatched again the next minute, and
the minute after, until the backlog drained. The next occurrence is now computed
from `now`.

`tick` also returns `coalesced`: the number of claims that absorbed at least one
further occurrence.

## Why

Every one of those replays is a real `stories.start` — a workspace provisioned
and an engine run charged to the project budget. An hourly schedule after a day
of worker downtime is **24 paid runs in 24 minutes**, against a budget sized for
one an hour, in a window narrow enough that each sees the last as still in
flight. A five-minute schedule is 288. Nothing bounded it: the replay was as long
as the outage. Downtime is not exotic — `tsx watch` restarts the worker on every
file change.

The replayed occurrences are also stale by construction. A security audit due
05:00 Monday and run 14:32 Tuesday is not the audit anyone asked for; it is the
same audit against a repository that has moved on, and the work it was meant to
precede already happened.

Computing from `now` makes the loop **level-triggered**: it converges on "this
schedule is due" and satisfies it once, however long the gap, rather than
replaying every edge it missed. A schedule claimed on time is unaffected, because
the next occurrence after `now` and the next after its own due instant are the
same one.

`due` counts rows, so a 24-occurrence catch-up and an ordinary minute were
indistinguishable in the worker log. `coalesced` costs one extra cron step per
due schedule — never a walk over the backlog — so an outage of any length is the
same work.

**What is deliberately untouched.** The claim stays a compare-and-swap on
`(nextRunAt, lastScheduledAt)`, so two workers still cannot take the same
occurrence. The dispatch still identifies itself by the occurrence it satisfies:
`lastScheduledAt`, the message dedupe key and `trigger.scheduledFor` all keep the
observed due instant rather than the wall clock, which is what keeps the dispatch
idempotent.

#330 offered three shapes — coalesce, bound the catch-up the way a Kubernetes
CronJob uses `startingDeadlineSeconds`, or make it the manifest's choice. This
implements coalescing, which is what every agent kickstart ships wants. It is
here as a concrete proposal, cheap to reject: if bounded catch-up is the intended
contract instead, the semantics live in one pure function and the test matrix
flips with it.

## Verification

The decision is extracted as a pure function, `scheduleAdvance`, so the semantics
are testable against a fixed clock with no database.

Unit — 3 of the 5 new cases are red against the previous behaviour, and the
on-time and clock-skew cases are green both ways because they must not regress:

```
× collapses a backlog of any size into a single claim
  AssertionError: expected 2026-09-06T07:00:00.000Z to deeply equal 2026-09-07T07:00:00.000Z
× reports a single missed occurrence, not only a long outage
  AssertionError: expected 2026-09-07T06:05:00.000Z to deeply equal 2026-09-07T06:10:00.000Z
× keeps the trigger timezone when catching up across a daylight-saving shift
  AssertionError: expected 2026-03-07T07:00:00.000Z to deeply equal 2026-03-10T06:00:00.000Z
```

Integration — `security-audit:nightly` put a week in arrears, then three ticks at
the same instant. Without the fix:

```
× runs a schedule that fell behind once, not once per missed occurrence
  AssertionError: expected 3 to be 1
```

Three ticks, three paid runs. With the fix, one run, one new story message, and
the schedule left ahead of the clock at `2026-01-11T02:00:00Z`.

```
vitest run test/agent-scheduler.test.ts test/agent-automation.integration.test.ts
                                                        21 passed
tsc --noEmit (@facility/api)                            clean
biome check (changed files)                             clean
```

The integration test asserts this project's own rows, not the tick counters:
`agent_schedules` is global and the shared test database accumulates rows across
suites, so a counter assertion there would pass or fail on history. I hit exactly
that while writing it. `scheduled` is safe because the fixture's catalog throws
for any other project, and the run was repeated three times consecutively to
confirm it.

identical 92/24 on a clean `main` — Windows noise in the patched `image-size`
cases, `tar` failing to resolve `C:`, and the registry publication tests. Beyond
the suite, the failure was reproduced end to end against a real PostgreSQL: three
ticks producing three dispatches before the change and one after.

Closes #330.

> 🤖 Claude Code helped

### DIFF
--- a/services/api/src/agents/scheduler.ts
+++ b/services/api/src/agents/scheduler.ts
@@ -36,9 +36,12 @@ export class AgentScheduler {
       .from(agentSchedules)
       .where(and(eq(agentSchedules.enabled, true), lte(agentSchedules.nextRunAt, now)));
     let scheduled = 0;
+    let coalesced = 0;
     for (const schedule of due) {
-      const claimed = await this.claim(schedule, now);
+      const advance = scheduleAdvance(schedule.cron, schedule.timezone, schedule.nextRunAt, now);
+      const claimed = await this.claim(schedule, now, advance.nextRunAt);
       if (!claimed) continue;
+      if (advance.missed) coalesced += 1;
       try {
         const [projectManifest, projection] = await Promise.all([
           this.projectManifests.load(schedule.orgId, schedule.projectId),
@@ -72,7 +75,7 @@ export class AgentScheduler {
         });
       }
     }
-    return { projects: activeProjects.length, due: due.length, scheduled, failures };
+    return { projects: activeProjects.length, due: due.length, scheduled, coalesced, failures };
   }
 
   async status(orgId: string, projectId: string) {
@@ -197,8 +200,7 @@ export class AgentScheduler {
     });
   }
 
-  private async claim(schedule: typeof agentSchedules.$inferSelect, now: Date) {
-    const nextRunAt = nextOccurrence(schedule.cron, schedule.timezone, schedule.nextRunAt);
+  private async claim(schedule: typeof agentSchedules.$inferSelect, now: Date, nextRunAt: Date) {
     return (
       await this.db
         .update(agentSchedules)
@@ -222,6 +224,31 @@ export class AgentScheduler {
 
 export function nextOccurrence(cron: string, timezone: string, from: Date) {
   return cronParser.parseExpression(cron, { currentDate: from, tz: timezone }).next().toDate();
+}
+
+/**
+ * Where a due schedule should point once its occurrence is claimed.
+ *
+ * The next occurrence is computed from `now`, not from the occurrence being
+ * claimed. Advancing by one cron step from the stored value makes the loop
+ * edge-triggered: a schedule that fell behind while the worker was down stays
+ * due after each claim and dispatches again on the following tick, one paid run
+ * per missed occurrence — 24 turns in 24 minutes for an hourly schedule after a
+ * day of downtime. Advancing from `now` makes it level-triggered: the loop
+ * converges on "this schedule is due" and runs it once, however long the gap.
+ *
+ * A schedule claimed on time is unaffected, because the next occurrence after
+ * `now` and the next occurrence after its own due instant are the same one.
+ */
+export function scheduleAdvance(cron: string, timezone: string, dueAt: Date, now: Date) {
+  const from = now > dueAt ? now : dueAt;
+  return {
+    nextRunAt: nextOccurrence(cron, timezone, from),
+    // Whether this claim absorbs more than the occurrence it satisfies: the
+    // following occurrence had already come due as well. One extra cron step,
+    // never a walk over the backlog, so an outage of any length costs the same.
+    missed: nextOccurrence(cron, timezone, dueAt) <= now,
+  };
 }
 
 function workspaceInput(manifest: ProjectManifest, defaultImage: string) {
--- a/services/api/test/agent-automation.integration.test.ts
+++ b/services/api/test/agent-automation.integration.test.ts
@@ -924,6 +924,60 @@ describe("agent automations use persistent story workspaces", async () => {
       await db.select().from(storyMessages).where(eq(storyMessages.storyId, scheduledStory.id)),
     ).toHaveLength(1);
   });
+
+  it("runs a schedule that fell behind once, not once per missed occurrence", async () => {
+    // security-audit:nightly is `0 2 * * *` UTC. Put it a week in arrears, the
+    // shape of a worker that was down: seven occurrences have come due.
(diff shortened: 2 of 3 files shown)


adrian-lorenzo (approved):
Thanks

Issue #325, opened 2026-09-07
Human authorization has no project dimension, but three docs pages say it does


Lob26:
Opening this as a question about the model rather than a bug report, because the fix is a design decision and I would rather hear the intent before anyone writes code against it.

## What the data model says

Machines have a project dimension. People do not.

- `apiKeys.projectId` exists, and `app.ts` enforces it well — a scoped key asking for another project gets **404, not 403**, with a comment saying exactly why. That is a careful piece of work.
- `orgMembers` is `(orgId, userId, roleId)`. `roles` is `(orgId, name, permissions[])`. There is no `projectMembers` table and no project column anywhere on the human path.
- Permission strings have no project dimension either: `PERMISSION_RESOURCES × {read, write}` plus `workspaces:execute`, `keys:issue`, `projects:kickstart`. `can(grants, needed)` is a string match with a `resource:*` wildcard.

So a role granting `workspaces:execute` grants it in **every** project in the organization. There is no way to express "can run stories on this one repository".

The bundled roles make the practical effect sharper:

```ts
{ name: "owner",      permissions: ["*"] },
{ name: "maintainer", permissions: ["*"] },   // identical
{ name: "viewer",     permissions: allReads },
```

`maintainer` and `owner` are the same grant. Between "read everything" and "administer the organization" there is nothing. Onboarding someone to run stories on one repository means handing them `roles:write`, `members:write`, `keys:issue` and `settings:write` across the whole org — a privilege escalation with no step in between. A custom role narrows the *verbs* but never the *project*.

## What the docs say

Three pages describe a project dimension that the code does not have:

- `apps/docs/docs/self-host/authentication.md` — "Any active **project maintainer** can read and continue the same story. Every project operation checks organization and project scope."
- `apps/docs/docs/self-host/authentication.md` — "each proxied request revalidates the user, **project membership**, workspace, service, expiry, and revocation status."
- `apps/docs/docs/contributors/architecture.md` — "resolves a principal, checks **project scope** and permissions."
- `apps/docs/docs/guides/operate-story.md` — "can continue it under the current user's **project membership**."

`README.md` also says "A signed-in **project member** can open an authenticated local preview."

The honest one is `reference/security.md`: *"Authorization uses organization membership, roles, and route permissions. Project-scoped API keys are pinned to one project."* That sentence is exactly right, and it is the only place the asymmetry shows.

[...]

PR #335, opened 2026-09-07
docs: state that human authorization is organization-wide

Lob26:
## What changes

Four documentation pages describe a project dimension on the human authorization
path that the data model does not have. They now say what the code does.

- `self-host/authentication.md` — "Any active **project maintainer** can read and
  continue the same story. Every project operation checks organization and
  **project scope**."
- `self-host/authentication.md` — the preview proxy "revalidates the user,
  **project membership**, ...". `WorkspacePreviewService.assertMembership`
  queries `org_members`; the check is organization membership.
- `contributors/architecture.md` — the request path "checks **project scope** and
  permissions".
- `guides/operate-story.md` — the UI "can continue it under the current user's
  **project membership**".

`reference/security.md` was already the accurate page. Its sentence becomes the
canonical one and now says plainly that a project boundary does not contain a
person, and that a project-scoped API key is the only mechanism that contains a
principal to one project.

Documentation only. No behaviour changes, and no position is taken on whether the
model should gain a project dimension — that question stays open in #325.

## Why

There is no `project_members` table. `roles` is `(orgId, name, permissions[])`
with no project column, and permission strings carry no project either — so a
role granting `workspaces:execute` grants it in **every** project in the
organization. "Maintainer on one repository" cannot be written down, so it cannot
be reviewed, audited, or refused.

This matters because the pages sit next to a genuinely careful piece of work: a
project-scoped API key asking for another project gets **404, not 403**, so it
cannot be used as an enumeration oracle. A reader who has just read that will
reasonably assume the same containment applies to people. It does not, and
nothing in the model made the asymmetry visible. Provisioning against the
stronger reading is a security expectation set by documentation, which is why
this half is worth fixing regardless of how the model question in #325 is
answered.

**Out of scope, deliberately.** #325 also notes that the bundled `owner` and
`maintainer` roles are both `["*"]` — identical grants under different names, with
nothing between "read everything" and "administer the organization". Narrowing
`maintainer` removes access from existing instances on deploy, so it belongs in
its own change with an expand → migrate → contract path, not in a documentation
correction. `reference/hardening.md` already recommends project-scoped keys for
automation that needs only one project, so no change was needed there.

One claim listed in #325 is already gone: the README no longer says a "project
member" can open an authenticated local preview. Nothing to correct there.

## Verification

Each claim was checked against the code before rewriting it, not against the
issue text:

```
grep projectMembers|project_members packages/db/src/schema.ts   → no matches
roles           = (id, orgId, name, description, permissions[])
orgMembers      = (id, orgId, userId, roleId)
BUNDLED_ROLES   → owner ["*"], maintainer ["*"], viewer allReads
preview.ts:221  → assertMembership(orgId, userId) selects from org_members
permissions.ts  → SPECIAL_PERMISSIONS includes "workspaces:execute"
```

```
```

`test:dev` reports **116 tests, 92 pass, 24 fail** here, and the identical 92/24
on a clean `main` — Windows noise unrelated to documentation. The meaningful
gates for a docs change are the link guard and the Docusaurus build, both green
above; the substance was verified by reading the schema, the role table and the
preview membership query rather than by any test.

Refs #325.

> 🤖 Claude Code helped

### DIFF
--- a/apps/docs/docs/contributors/architecture.md
+++ b/apps/docs/docs/contributors/architecture.md
@@ -47,7 +47,8 @@ similar handwritten versions.
 
 ## Main request path
 
-A UI or MCP request enters the API, resolves a principal, checks project scope and permissions, and
+A UI or MCP request enters the API, resolves a principal, checks its organization-wide role
+permissions and — for a project-scoped API key — that the key's project matches the request, and
 calls a domain service. Starting or continuing a story persists its message and turn in PostgreSQL.
 The worker claims the turn, resolves an exact agent manifest, wakes the workspace provider,
 prepares `.facility.yml`, issues repository credentials, and starts or resumes Claude Code or
--- a/apps/docs/docs/guides/operate-story.md
+++ b/apps/docs/docs/guides/operate-story.md
@@ -37,7 +37,7 @@ Messages sent while a turn is active wait in order.
 
 Use `facility_get_story` to inspect status and `next_operations`. Use
 `facility_get_conversation` with its cursor for durable message history. The UI renders the same
-conversation and can continue it under the current user's project membership.
+conversation and can continue it under the current user's organization role.
 
 The story timeline is the review path across the whole delivery. It shows which agent, model,
 session, workspace, branch, and initial SHA started each turn; the final SHA, commits, files, and
--- a/apps/docs/docs/reference/security.md
+++ b/apps/docs/docs/reference/security.md
@@ -37,10 +37,12 @@ Facility still enforces:
 - secret redaction from persisted command events and API responses; and
 - explicit confirmation and idempotency for durable workspace deletion.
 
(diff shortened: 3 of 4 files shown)


adrian-lorenzo (changes requested):
Thanks for clarifying the organization-wide authorization model and updating the newer story docs.

One detail to correct: reading a story or conversation requires `projects:read`; `workspaces:execute` permits continuing it and does not imply read access. Please name both permissions in the authentication guide.

Lob26:
Addressed at 9e5ccd8. You are right, and naming only one of them was the same failure this PR set out to fix: prose implying a containment the model does not have. The guide now says `projects:read` reads a story and its conversation, `workspaces:execute` continues it, opens its preview or suspends it, and that execute does not imply read, so a role needs both. Checked against every `app.get` in `routes/v1/story-workspaces.ts` rather than inferred. Thanks for catching it.

Issue #331, opened 2026-09-07
Web UI accessibility: timeline has no live regions, no skip link, no sr-only primitive, no a11y guard in CI


anabelle:
### The problem

Facility's operating model makes the web UI the surface where humans supervise agents: read the story timeline, review turns, decide. Parts of that surface are currently hard or impossible to use with a keyboard or screen reader. Findings below come from a static pass over `apps/web` and `packages/ui` plus a runtime check of the login page, at commit c14bc6d.

Worth saying first, because the audit found it: the fundamentals are better than most applications I audit. `apps/web/components/shell/nav.tsx:52` has `<nav aria-label="Primary">` with a proper `aside` landmark, `packages/ui/src/tokens.css:87` applies a global `prefers-reduced-motion` kill-switch, status is always text plus color (`apps/web/components/ci-status.tsx` renders "checks · failed · …", story rows render `story.status` as text), the forms have correct live regions (`apps/web/components/settings/keys-manager.tsx:80` uses polite/assertive correctly), the command palette is a labelled dialog with Escape support (`apps/web/components/shell/cmdk.tsx:128`), and there are no clickable-`div` antipatterns.

The gaps:

1. **No skip link anywhere.** The sidebar navigation is tabbed through on every page (WCAG 2.4.1 bypass blocks).
2. **No `sr-only` primitive in `packages/ui`.** Icon-only controls have no idiomatic way to get an accessible name; `aria-label` coverage across the app is sparse.
3. **The story timeline, the core review surface, has no live-region semantics.** Live regions exist only in settings and forms components. Turn state changes and completions are visual only — `StatusDot` is `aria-hidden` by design and the adjacent status text is never announced. A screen-reader reviewer cannot tell when an agent turn finishes, which is the exact moment the product asks them to pay attention.
4. **No automated accessibility checking.** No axe-core anywhere in the test suites, and biome runs the default preset only. Nothing fails if any of the above regresses further.
5. **The login page has no heading and no `<main>` landmark**, and the document title is bare "facility".
6. **Review surfaces rely on 9–12.5px muted text** (`LegendChip` is 9px, story metadata 10.5px with `--dim`). Contrast at those sizes and 200% zoom behavior are unverified.

### What I propose

1. Treat accessibility as a guard, matching the repository's own invariants approach: axe-core in the web app's vitest suite, deterministic (jsdom, server-rendered components, no network, no credentials), covering the `packages/ui` primitives and key stateless `apps/web` components. The guard set only grows.
[...]

PR #337, opened 2026-09-07
feat(web): baseline accessibility semantics for the shells and login

anabelle:
Refs #331. The baseline-semantics slice agreed in the issue: no live-region decisions here, no new dependencies.

## What changes

- `VisuallyHidden` primitive in `@facility/ui`: the blessed way to give icon-only controls an accessible name. Tailwind's `sr-only` technique, never `display: none`, so content stays in the accessibility tree.
- `SkipLink` in both authenticated shells (org and project): first focusable element in the DOM, `sr-only` until focused, fixed top-left when visible, targets an explicit `#main-content` with `tabIndex={-1}` so focus actually moves and the next Tab lands inside the work area instead of back in the sidebar.
- Login page: `<main>` landmark, the brand mark becomes the `<h1>`, and `title: "Sign in"` rides the root template to render "Sign in · facility".

## Tests

`apps/web/test/a11y-baseline.test.tsx`, deterministic and dependency-free: server-renders both real layouts with the API, topbar, and router-link modules mocked, then asserts the DOM contract (skip link precedes the sidebar nav, `main` carries `id` and `tabindex`, login has landmark, heading, and title; `VisuallyHidden` keeps text in the tree). Runs in the default node environment, no browser, no network, no credentials.

## Verification

- `pnpm --filter @facility/web test` - 6 files, 37 tests passed (5 new)
- `pnpm --filter @facility/web typecheck` - clean
- `pnpm --filter @facility/ui typecheck` - clean
- `pnpm exec biome check` on changed files - clean after import-order fixes
- `pnpm --filter @facility/web build` - clean
- `node guards/run.mjs` - 2 guards ran, 0 failed

Not run: full `pnpm verify` (needs Docker-backed PostgreSQL; no API, database, or workspace boundary is touched) and the manual keyboard pass (recommended on merge: tab from the address bar on an org page, a project page, and the login page).

## Effects

- User-visible: keyboard and screen-reader users can bypass the sidebar on every authenticated page; the login page exposes a landmark, a heading, and a descriptive title. `feat` on `0.x` releases a patch.
- No persistence, migration, or compatibility effect. No security or money surface touched; no dependencies added.
- Follow-ups stay in #331: the story-timeline announcement strategy (design note promised in the issue) and folding these surfaces into the #332 guard once it lands.

### DIFF
--- a/apps/web/app/(app)/(org)/layout.tsx
+++ b/apps/web/app/(app)/(org)/layout.tsx
@@ -1,4 +1,5 @@
 import { MobileNav, Sidebar } from "@/components/shell/nav";
+import { SkipLink } from "@/components/shell/skip-link";
 import { Topbar } from "@/components/shell/topbar";
 import { api } from "@/lib/api";
 
@@ -8,11 +9,14 @@ export default async function OrgLayout({ children }: { children: React.ReactNod
 
   return (
     <div className="flex min-h-dvh">
+      <SkipLink />
       <Sidebar />
       <div className="min-w-0 flex-1">
         <MobileNav />
         {me.ok ? <Topbar me={me.data} projects={projectList} /> : null}
-        <main className="px-5 py-8 sm:px-8 lg:px-10">{children}</main>
+        <main id="main-content" tabIndex={-1} className="px-5 py-8 sm:px-8 lg:px-10">
+          {children}
+        </main>
       </div>
     </div>
   );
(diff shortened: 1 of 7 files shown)


adrian-lorenzo (commented):
Thanks for improving accessibility and covering both app shells in the tests.

This overlaps with #339, which we’ve already reviewed positively. Could you coordinate on a single PR for this slice, keeping the shell-level regression tests from this one?

PR #340, opened 2026-09-08
fix: escape untrusted YAML values in generated workspace templates

o-mid:
Addresses #228.

Facility 0.12 no longer generates GitHub workflows or Claude settings, so the original doctor-watch, settings.json, and protect-branch regex vectors are gone. Model ids were still interpolated unquoted into agent YAML frontmatter, so values like a newline plus `enabled: false`, `|`, `#`, or `$(id)` could change structure or truncate the scalar.

This quotes every generated YAML scalar (models and environment commands) in both `facility init` and kickstart rendering, and adds hostile-input regressions through the CLI, core renderer, and agent parser.

## Test plan
- [x] `pnpm test --filter @theagilemonkeys/facility`
- [x] `pnpm test --filter @facility/core`
- [x] `pnpm test --filter @facility/agents`
- [x] Confirm hostile model ids (`$(id)`, `foo"bar # pwned`, newline + `enabled: false`, `|`) stay quoted scalars and parse as the original model string
- [x] Confirm ordinary ids (`gpt-5.6-sol`, dated `claude-opus-4-8-20260101`) still work after quoting
- [x] Confirm `setup` / `start` / `ready` commands with quotes and `$(id)` remain data, not shell

### DIFF
--- a/packages/agents/test/agents.test.ts
+++ b/packages/agents/test/agents.test.ts
@@ -200,10 +200,10 @@ describe("agent manifests", () => {
         .map(async (file) => ({
           file,
           source: (await readFile(join(templateDirectory, file), "utf8"))
-            .replaceAll("{{PLAN_MODEL}}", "claude-opus-4-8")
-            .replaceAll("{{REVIEW_MODEL}}", "claude-sonnet-4-6")
-            .replaceAll("{{CODEX_BUILD_MODEL}}", "gpt-5.6-sol")
-            .replaceAll("{{CODEX_PLAN_MODEL}}", "gpt-5.6-sol"),
+            .replaceAll("{{PLAN_MODEL}}", JSON.stringify("claude-opus-4-8"))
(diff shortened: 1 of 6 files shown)


adrian-lorenzo (changes requested):
Thanks for adapting this to the current workspace templates.

CLI substitution still interprets values such as `$&` and `{{CODEX_PLAN_MODEL}}`, corrupting generated models. The doctor change also accepts invalid model values. Please use the core renderer’s single-pass callback approach, preserve model validation, and add regression tests for both.

o-mid:
@adrian-lorenzo, 
CLI init used String.replaceAll, so $& became the matched placeholder and a model value of {{CODEX_PLAN_MODEL}} got substituted again when that key was applied. It now uses the same single-pass callback as the core renderer.

Doctor parses model as a YAML string (1–160 chars). Comment-only lines and unclosed quotes fail; quoted names with spaces still pass.

Added regressions for both in the CLI, and the $& / nested placeholder case in core.

PR #342, opened 2026-09-08
fix: scope facility doctor agent checks to YAML frontmatter

o-mid:
Addresses doctor false pass and false fail against the 0.12 agent contract.

`facility doctor` applied name, model, trigger, and forbidden-key regexes to the whole agent file. Prompt text such as a GitHub `permissions:` example failed the check, and a wrong frontmatter `name:` could be hidden by repeating the expected name in the prompt. Quoted model ids with spaces also failed.

Doctor now splits YAML frontmatter from the prompt, accepts `mcp` and `ui` triggers (already valid at runtime), and reads the model as a YAML string scalar. Comment-only values (`model: # choose a model`) and unclosed quotes are rejected, matching the server.

## Test plan
- [x] `pnpm --filter @theagilemonkeys/facility test`
- [x] CLI suite 10 consecutive passes
- [x] Confirm a prompt that contains `permissions:` still passes
- [x] Confirm frontmatter `permissions:` still fails
- [x] Confirm `name: builder` only in the prompt does not satisfy builder.md
- [x] Confirm `model: "gpt-5.6-sol with spaces"` passes
- [x] Confirm `model: # choose a model` fails
- [x] Confirm `model: "unclosed` fails
- [x] Confirm an agent with only `mcp` and `ui` triggers passes

### DIFF
--- a/packages/cli/src/doctor.mjs
+++ b/packages/cli/src/doctor.mjs
@@ -53,15 +53,32 @@ function checkAgent(dir, name) {
   const path = join(dir, relative);
   if (!existsSync(path)) return failed(relative, "missing");
   const source = readFileSync(path, "utf8").replace(/\r\n?/g, "\n");
-  if (!source.startsWith("---\n") || !/\n---\n[\s\S]*\S/.test(source)) return failed(relative, "invalid frontmatter or empty prompt");
-  if (!new RegExp(`^name:\\s*${escapeRegExp(name)}\\s*$`, "m").test(source)) return failed(relative, `name must be ${name}`);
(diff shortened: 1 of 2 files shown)


adrian-lorenzo (changes requested):
Thanks for fixing the frontmatter scoping.

The relaxed model check now accepts `model: # choose a model` and unclosed quoted values, although the server rejects both. Please validate the model value while preserving quoted names with spaces, and add regression tests for these cases.

o-mid:
@adrian-lorenzo Thanks, that was a real hole.
 \S treated # and an opening quote as a model ID.
Follow-up parses the model as a YAML string scalar, so quoted names with spaces still pass. model: # choose a model and unclosed quotes fail, same as the server. Tests cover both.

Issue #284, opened 2026-09-02
[0.12] Prove migration, retention, and release readiness


adrian-lorenzo:
## Goal

Prove the complete Facility 0.12 workflow and prepare the breaking release.

## Automated journey

Add an end-to-end command that uses real application boundaries and deterministic local fakes. It should:

1. run kickstart and inspect the generated `.agents/` crew;
2. create a project and story through the actual MCP endpoint;
3. select an agent and verify its configured model reaches a fake engine;
4. create a workspace, worktree, and native session marker;
5. suspend and replace compute, then resume the same files;
6. switch agents or engines without deleting the earlier session;
7. start the Compose fixture and run its browser test;
8. open the authenticated preview through the local proxy;
9. create a branch, commit, push, issue comment, workflow change, and pull request against a GitHub fake;
10. dispatch PR reviewer, CI doctor, and a scheduled security audit;
11. merge, archive, restore, and confirm that durable state remains;
12. rerun the story checks from the UI issue through a real browser;
13. request explicit deletion and verify isolated, idempotent cleanup.

The default suite uses local servers, local repositories, a fake clock, and fake Claude Code and Codex executables. It cannot require internet access or live credentials.

## Retention tests

Run a 14-calendar-day soak with a merged and archived story. The release waits for this test to finish. The worktree and native session files must still be present after wake.

Re-run the backup, restore, provider failure, preview, compute replacement, and explicit deletion checks implemented by the architecture and removal issues against the release candidate.

## Pilot work

Before release, use one authorized repository with Claude Code and one with Codex. Run an interactive story, a PR event, a CI doctor event, and a scheduled security audit. Test the preview from another computer or network.

Record defects as separate issues under the epic when they can be fixed independently.

## Documentation and release

Update the README, quickstart, architecture, security model, MCP reference, UI guide, self-hosting guide, migration guide, and roadmap. The docs should describe `.agents/`, full maintainer access, indefinite state retention, scheduled agents, and explicit deletion in direct language.

Publish the release with a breaking Conventional Commit subject, for example:

`feat!: replace governed runs with persistent story workspaces`

The pull request title and commit range must declare the same release impact.

## Acceptance criteria

- [ ] The full automated journey passes from a clean checkout.
- [ ] Security-sensitive tests cover valid, malformed, expired, revoked, replayed, and cross-tenant inputs.
- [ ] The 14-calendar-day retention test keeps sessions and worktree files.
- [ ] Backup and restore recover the same workspace data.
[...]

PR #343, opened 2026-09-08
fix(web): align story creation permission and verify workspace lifecycle

shuntianyifang:
The Stories page checks the retired `runs:execute` permission before offering story creation, while the workspace API requires `workspaces:execute`. A role with workspace execution permission can create a story through the API but cannot start one from the page; a role with only the old permission is offered an action the API rejects.

This change aligns the page with the API and adds browser regression coverage for the persistent workspace lifecycle. Refs #284; this is a bounded UI acceptance contribution, not completion of the release-readiness issue.

## Changes

- Check `workspaces:execute` when rendering story creation, with six permission regression cases covering the exact permission, wildcards, read-only permissions, and the retired permission.
- Add four Chromium tests against the real Next app and a loopback HTTP fixture: creation, suspend/continue, archive/restore, retained history, read-only controls, deletion confirmation/cancellation, visible errors and retry, and no restore after deletion.
- Add one Chromium test using the real API, session authorization, PostgreSQL and `FakeWorkspaceRuntime`. It verifies the same workspace and untracked file/Claude/Codex session-marker contents survive lifecycle operations, then verifies confirmed deletion removes the files while retaining conversation history.
- Run both browser suites in CI after `pnpm verify`; document the setup and fixture boundaries. Keep Playwright files outside Vitest discovery.

## Latest acceptance

Merged main `c14bc6d`; the PR head is now `daa177de4b6a919cc2e6e05818311d7f4ee5bb85`. [Ubuntu acceptance run 34309420909](https://github.com/shuntianyifang/facility/actions/runs/34309420909) explicitly checks out that exact source tree, without the auxiliary validation workflow in the tested checkout.

- Frozen install, Terraform fmt/init/validate/test, and migration compatibility passed.
- `pnpm verify` passed its lint, type checks, clean builds, critical integration suites, remaining tests, unused checks and guards, but **failed the final dependency audit**: 18 findings, including 2 critical and 6 high (2 already ignored by repository policy). No exclusions were added. The affected existing versions, including Next.js 16.2.11, are unchanged by this PR. The full verify job is not green.
- Chromium: **4 UI tests passed** (30.5s), plus **1 real API/PostgreSQL/filesystem browser test passed** (26.9s). The auxiliary workflow collected this evidence after audit failure while preserving the failed job result.
- Node 22.13.0 compatibility job passed.
- Original Compose and Vercel runner image builds passed. Container smoke checks passed with `FACILITY_PREVIEW_URL=https://preview.example.test`: migrations exit 0, API readiness, login page, unauthenticated `/v1/me` rejection, and API/Web restart recovery. Disposable containers and volumes were removed afterward. This does not validate external TLS or preview traffic.
- Direct Edge browser interaction now passed against the real local Next app and disposable loopback fixture: create, suspend, send/resume, archive, reload, restore, deletion confirmation/cancellation, injected failure and retry, retained history/no restore after deletion, and read-only controls. This was a step-by-step browser-control pass, in addition to the automated suites.

## Remaining blockers and boundaries

- Dependency audit must be resolved before claiming full acceptance. Previously passing audit results are historical and do not supersede the current failure.
- Default Compose startup fails production validation because its preview URL is HTTP. This was reproduced in [run 34307866659](https://github.com/shuntianyifang/facility/actions/runs/34307866659); existing #333 addresses it. The configured smoke pass above does not claim the default configuration works.
- Upstream PR CI and independent review have no results at the time of this update; keep this PR in draft.
- These fixtures do not prove real hosted engine resume, compute replacement, backup recovery or the 14-day retention/pilot work in #284. Docker-backed workspace E2E remains a policy-allowed skip for this PR's paths.

## Scope and compatibility

This is a user-visible permission-alignment fix, classified as `fix(web)` (patch, no breaking change). Server-side authorization is unchanged. No migrations, persistence format, budget/cost enforcement or production observability behavior changes.

External catalog/manifest reads and execution use deterministic fakes; tests need no live credentials or network after dependencies and Chromium are installed. The real-API fixture binds to loopback and refuses databases other than local `facility_test`/`facility_ws`.

Checked existing PRs on 2026-09-08: #332 covers axe checks on rendered fragments, #337/#339 cover accessibility semantics, and #324 covers Docker wake readiness. Those do not supply this lifecycle browser suite or the story-creation permission fix.


adrian-lorenzo (approved):
Thanks for adding this coverage.

Verified the lifecycle flows, including file retention and confirmed deletion through the real API. No blocking issues.

PR #356, opened 2026-09-09
fix(api): keep artifacts when browser tests fail

### LINKED ISSUE #352: Failed browser tests leave screenshots and traces out of the story
## What happened

When a browser test fails, its screenshots and traces can still be on disk, but they don't appear in the story's artifacts.

`runBrowserTest` throws as soon as the command exits with a nonzero status. That skips the code that finds the files in `FACILITY_ARTIFACT_DIR`, registers them, and records the browser-test event. I'd expect Facility to save that evidence and still report the test failure.

## How to reproduce

1. Set `environment.browser_test` to write a screenshot or trace into `$FACILITY_ARTIFACT_DIR` and then exit with a nonzero status.
2. Prepare the story workspace and run its browser test.
3. Check the story's artifacts and workspace events.

The test failure is reported, but the files it wrote aren't registered and the browser-test event is missing.

## Version

## Evidence

[Failure precedes artifact discovery and persistence](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/services/api/src/workspaces/project-environment.ts#L384-L425). The regression test uses disposable PostgreSQL and real files through the fake workspace runtime. The real Docker workspace check is still blocked by a runner architecture mismatch.

TheNaubit:
## What changes

Save the screenshots, traces and reports a browser test writes, even when the test exits with an error. The service registers those files and records the redacted result event before returning the failure.

If saving the evidence also fails, keep the original test error so it doesn't get hidden by the second failure. The docs explain how to retrieve failed-test artifacts from the story, since a failed operation doesn't return the successful result's artifact list.

This is a nonbreaking fix using the existing artifact and event records. It needs no migration and doesn't change permissions, budgets or the API schema.

## Why

Closes https://github.com/theam/facility/issues/352

When a browser test fails, the screenshots and traces are often what you need to work out why. Throwing before registering them leaves that evidence out of the story.

## Verification

- `pnpm --filter @facility/api exec vitest run test/project-environment.integration.test.ts --fileParallelism=false` passed all 10 tests. The new cases check that failed tests keep their screenshot and trace, secrets are redacted, and an artifact scan error doesn't replace the original test error.
- API typecheck, Biome checks, docs tests and the docs build passed.
- `pnpm verify` passed. The dependency audit still has the repo's two existing ignored high advisories; this PR adds no exceptions.

- [ ] Behaviour verified beyond the test suite (say how)

### DIFF
--- a/apps/docs/docs/reference/project-manifest.md
+++ b/apps/docs/docs/reference/project-manifest.md
@@ -127,8 +127,9 @@ Facility's preview handoff provides authenticated browser access.
 ## Browser test artifacts
 
 When `browser_test` runs, Facility sets `FACILITY_ARTIFACT_DIR`. Write screenshots, traces, logs,
-and reports beneath that directory. Facility records the resulting files as story artifacts and
-returns their identifiers and URIs with the operation result.
+and reports beneath that directory. Facility records the resulting files as story artifacts,
+including files written before a failed command. Failed-test artifacts remain on the story timeline
+for retrieval afterward; successful operations also return their identifiers and URIs.
 
 ## Validation
 
--- a/services/api/src/workspaces/project-environment.ts
+++ b/services/api/src/workspaces/project-environment.ts
@@ -386,42 +386,57 @@ export class ProjectEnvironmentService {
       preparedInput.credentials.environment,
       preparedInput.manifest.environment.secrets,
     );
-    if (result.exitCode !== 0) throw commandFailure("environment.browser_test", script, safeResult);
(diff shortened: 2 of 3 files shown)


adrian-lorenzo (changes requested):
Thanks for fixing this.

I confirmed the timeout case from the existing review: a screenshot written before timeout remains on disk but never reaches the story. Please collect artifacts when execution throws too, preserve the original error, and add a timeout regression test.

PR #357, opened 2026-09-09
fix(workspaces): enforce readiness command deadlines

### LINKED ISSUE #353: Readiness checks can run past the two-minute timeout
## What happened

A readiness check can keep a preview or browser test waiting well past the two-minute deadline.

The polling loop checks the deadline between attempts, but each `environment.ready` command gets the usual 30-minute command timeout. It also accepts a successful result that arrives after the deadline. The first check when reopening a prepared workspace has the same problem.

I'd expect the deadline to stop a hanging check, including that first check.

## How to reproduce

1. Set `environment.ready` to `sleep 180`.
2. Start the environment or reopen a prepared workspace.
3. Wait for the readiness check.

The command can keep running beyond the two-minute deadline and still be accepted when it exits successfully.

## Version

## Evidence

[Readiness loop and generic command timeout](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/services/api/src/workspaces/project-environment.ts#L526-L574), [prepared-workspace probe](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/services/api/src/workspaces/project-environment.ts#L285-L294). I traced this in the source. The regression test uses a shorter deadline and a real sleeping subprocess; I haven't completed the real Docker workspace check because the runner won't start in the ARM sandbox.

TheNaubit:
## What changes

Give each readiness command only the time left before the deadline. This also covers the first check when reopening a prepared workspace, and rejects a successful result if it arrives too late.

A runtime timeout becomes `environment_not_ready`; other runtime errors keep their existing behavior. Setup and start commands keep their longer timeout.

This is a nonbreaking timeout fix with no migration or permission/budget changes. Timing out a readiness check keeps the prepared files and data and doesn't rerun the seed step.

## Why

Closes https://github.com/theam/facility/issues/353

The two-minute deadline only controlled when the loop started another check. One hanging command could still make the operation wait for up to 30 minutes.

## Verification

- `pnpm --filter @facility/api exec vitest run test/project-readiness.test.ts test/project-environment.integration.test.ts --fileParallelism=false` passed all 15 tests.
- The tests cover retries with less time remaining, late success, timeout errors, healthy workspace reuse and unrelated runtime errors. The integration test stops a real sleeping command and checks that the setup and seed data survive.
- API typecheck and `pnpm verify` passed.

- [ ] Behaviour verified beyond the test suite (say how)

### DIFF
--- a/apps/docs/docs/reference/project-manifest.md
+++ b/apps/docs/docs/reference/project-manifest.md
@@ -89,7 +89,9 @@ setup so a normal wake does not repeatedly seed the environment.
 
 Make `start` safe to repeat. It should return after bringing services up; use a detached Compose
 command, process manager, or equivalent. Make `ready` fail quickly while the service is unavailable
-and succeed only when a user can exercise it.
+and succeed only when a user can exercise it. Readiness checks share a two-minute deadline,
+including time spent inside each check. Reopening a prepared workspace also bounds its initial
+readiness probe; a timed-out probe does not restart or reseed the application.
 
 ## Secrets and variables
 
--- a/services/api/src/workspaces/project-environment.ts
+++ b/services/api/src/workspaces/project-environment.ts
@@ -19,6 +19,7 @@ import type {
   WorkspaceLocator,
   WorkspaceRuntime,
 } from "./runtime.js";
+import { WorkspaceRuntimeError } from "./runtime.js";
 
 const RepositoryName = z
   .string()
@@ -288,7 +289,7 @@ export class ProjectEnvironmentService {
     const preparedInput = await this.withDeclaredEnvironment(input);
     const ready = preparedInput.manifest.environment.ready;
     const alreadyReady = ready
-      ? (await this.command(preparedInput, ready, primaryPath(preparedInput.credentials)))
+      ? (await this.checkReady(preparedInput, Date.now() + (input.readinessTimeoutMs ?? 120_000)))
           .exitCode === 0
       : false;
     return this.startServices(preparedInput, input.setupChecksum, alreadyReady);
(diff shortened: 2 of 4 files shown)


adrian-lorenzo (changes requested):
Thanks for fixing this.

I confirmed the Docker issue noted in the existing review: a 100 ms readiness deadline still waited about 2 seconds for `sleep 2` to finish. Please make the timeout settle independently of stream closure, ensure the command is terminated, and add regression coverage.

Lob26:
On the Docker behaviour adrian flagged, that a 100 ms deadline still waited about 2 s for `sleep 2`: I think the cause is in the runtime rather than in your deadline arithmetic, and it is one line.

`DockerWorkspaceRuntime.exec` settles on stream closure, not on the timer:

```ts
const timeout = command.timeoutMs
  ? setTimeout(() => { timedOut = true; void this.terminateExecution(container, execution); }, command.timeoutMs)
  : undefined;
try {
  await waitForStream(stream);      // <- the promise waits here
  ...
  if (timedOut) throw new WorkspaceRuntimeError("workspace_command_timeout", ...);
```

The timer only sets a flag and fires `terminateExecution`, which resolves the exec's pid and sends `kill -TERM` through a second exec. So `await exec(...)` resolves when the attached stream closes, and `timedOut` is read after that. Two things stretch it: the round trip to inspect and start the killer, and `SIGTERM` reaching only the pid of the exec's own process, so a shell that spawned a child keeps the stream open until the child exits.

That is why passing a smaller deadline down does not shorten the wall clock. Making the timeout settle independently, the way adrian asked, probably means racing `waitForStream` against the timer in `exec` and treating the terminate as best-effort cleanup, rather than changing anything in `project-environment.ts`.

Worth flagging that this touches the same path as #324, which changes the workspace bootstrap and the readiness probe in `docker.ts`. Neither of us can see the other side from our own branch, so whichever lands second will want a look at the merge.

I have the Docker side loaded from that work, so if you want a patch for the `exec` race against your branch, say the word. Equally happy to leave it entirely with you.

PR #358, opened 2026-09-09
fix(cli): use the server manifest validation in doctor

### LINKED ISSUE #354: facility doctor accepts manifests that the server rejects
## What happened

`facility doctor` can say a `.facility.yml` file is valid even though the server will reject it when creating a workspace.

Doctor checks a few lines with regular expressions. It doesn't parse the full YAML or validate its structure against the server's schema. For example, adding an unsupported field under `environment` passes the local check.

I'd expect doctor to catch this before I try to create a workspace.

## How to reproduce

1. Run `facility init` with a valid repository and start command to generate `.facility.yml` and the agent files.
2. Add `unexpected: true` under `environment` in `.facility.yml`.
3. Run `facility doctor --json`.

Doctor accepts the manifest, but the server rejects the unknown field. Duplicate YAML keys and malformed structures can cause similar disagreements.

## Version

## Evidence

[CLI checks](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/packages/cli/src/doctor.mjs#L35-L48), [server manifest contract](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/services/api/src/workspaces/project-environment.ts#L25-L109). PR #342 addresses agent frontmatter, which is a separate validation path.

TheNaubit:
## What changes

Have `facility doctor` and the API use the same YAML parser and Zod schema for `.facility.yml`, so doctor catches manifests the server would reject.

The shared parser stays in the published CLI package. The API imports it, derives its types from it, and includes the file in Turbo's cache inputs so parser changes invalidate cached builds. The API keeps its existing error code and manifest hash behavior.

The CLI adds `yaml` and `zod`, which the API already uses and the lockfile already contains. This is a nonbreaking validation fix: invalid manifests fail earlier, with no data migration or changes to permissions or budgets.

## Why

Closes https://github.com/theam/facility/issues/354

A successful doctor check should mean the manifest passes the server's validation. The current regular expressions let unsupported fields and malformed structures through, so users only find out later when they try to create a workspace.

## Verification

- All 11 CLI tests and both API manifest tests passed, including checks for generated manifests, unknown fields, duplicate keys and malformed YAML.
- I installed the packed CLI into a fresh temporary fixture, ran `init`, and checked it with `doctor --json`. It accepted the generated manifest and rejected it after I added an unknown environment field.
- API typecheck, the production build and declaration build passed. Changing the shared parser changed the API's Turbo task hash as expected.
- The production API Docker image built successfully, including the CLI entrypoint/import check. The packed CLI includes the parser.
- `pnpm verify` passed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -97,8 +97,8 @@ RUN test -f /app/node_modules/@facility/db/dist/bin/deploy.js \
 # tasks inside the VPC. `facility instance bootstrap` needs the database, and in
 # a reference deployment the database accepts connections only from the service
 # security group — there is nowhere else to run it from. The package is a few
-# hundred kilobytes of plain ESM whose only dependency, `postgres`, is already
-# here for @facility/db.
+# hundred kilobytes of plain ESM whose production dependencies are already
+# installed for the API image.
 COPY --from=build-api /app/packages/cli /app/cli
 # Operator commands read as `facility …` inside the image, the way they read
 # everywhere else, instead of as a path into it. A wrapper rather than a symlink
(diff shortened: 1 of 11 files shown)


adrian-lorenzo (approved):
Thanks for fixing this.

Verified the shared validation and packaged CLI behavior; no blocking issues.

PR #359, opened 2026-09-09
fix(web): restore command palette navigation

### LINKED ISSUE #355: Command+K does nothing, and the palette lists pages that no longer exist
## What happened

The topbar shows a Command+K shortcut, but pressing it doesn't open anything.

Neither the project layout nor the organization layout mounts the existing command palette. While checking that, I also found that the component still lists removed sessions, approvals, harness and audit pages.

The shortcut should open the palette, and its entries should lead to pages that exist.

## How to reproduce

1. Open a project or organization page.
2. Press Command+K on macOS, or Ctrl+K.
3. Notice that no palette opens, despite the shortcut in the topbar.

The stale destinations are in the unused palette component; they aren't reachable through the shortcut while this mounting bug is present.

## Version

## Evidence

[Advertised shortcut](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/apps/web/components/shell/topbar.tsx#L24), [obsolete entries](https://github.com/theam/Facility/blob/6338c8c4259b63df50efb28b904907ecc3388e94/apps/web/components/shell/cmdk.tsx#L44-L68). I reproduced the broken shortcut in the production build, using a local API fixture with fixed responses.

TheNaubit:
## What changes

Mount the existing command palette in both the project and organization layouts so Command+K and Ctrl+K open it.

Use the sidebar's project navigation list for the palette entries and remove the organization links to pages that no longer exist. Filtering, arrow-key selection and project switching still work. The usage guide now documents the shortcut.

This is a nonbreaking UI fix. It adds no dependencies and doesn't change stored data, APIs, permissions or budgets.

## Why

Closes https://github.com/theam/facility/issues/355

The topbar shows a shortcut that does nothing because the palette is never mounted. Its old destinations also need updating before people can use it.

## Verification

- All six Vitest/jsdom tests passed. They cover both layouts, existing routes, filtering, keyboard selection, organization navigation and project switching.
- Web typecheck, the production build and `pnpm verify` passed.
- I also checked the production build in a browser with a local API fixture that returns fixed responses. All 16 checks passed, including opening and dismissing the palette, keyboard navigation, project switching and organization pages. There were no browser errors.

### DIFF
--- a/apps/docs/docs/guides/operate-story.md
+++ b/apps/docs/docs/guides/operate-story.md
@@ -8,6 +8,9 @@ MCP is the primary way to automate a Facility story. The web UI exposes the same
 useful for browsing conversations, inspecting environments, opening previews, and taking lifecycle
 actions.
 
+In the web UI, press Command+K on macOS or Ctrl+K to open navigation. Type to filter projects and
+sections, use the arrow keys and Enter to choose a destination, or press Escape to dismiss it.
+
 ## Start a story
 
 Before starting, identify the project, the agent, and the body of work. For GitHub work, use a
--- a/apps/web/app/(app)/(org)/layout.tsx
+++ b/apps/web/app/(app)/(org)/layout.tsx
@@ -1,3 +1,4 @@
+import { CommandPalette } from "@/components/shell/cmdk";
 import { MobileNav, Sidebar } from "@/components/shell/nav";
 import { Topbar } from "@/components/shell/topbar";
 import { api } from "@/lib/api";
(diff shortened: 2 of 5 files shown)


Lob26:
Heads-up: this branch predates #368, which dropped Pipeline from `projectNav`. Merged onto current `main` (b7a9212), `apps/docs/docs/guides/operate-story.md` conflicts and 2/6 tests in `apps/web/test/command-palette.test.tsx` fail:

- the sections loop still expects `alpha · pipeline` (L94–95);
- the ArrowDown×2 case now lands on `/projects/project-a/insights` instead of `/projects/project-a/pipeline` (L114).

The component change itself holds up — it reads `projectNav`, so it followed #368 on its own — so this should just be a rebase plus those two expectations.

PR #365, opened 2026-09-10
fix(turns): carry the asked question into a rebuilt prompt

### LINKED ISSUE #344: Rebuilt prompt omits the blocking question when its reply goes to a fresh agent
## What happened

At `c14bc6db34862e2e3135f9f6c0cf44370a998df8`, a terminal `<facility-needs-attention>` question is extracted into attention detail and removed from the agent's ordinary conversation message. A subsequent reconstructed prompt includes the human reply but does not include that attention detail.

This matters when the reply goes to a different agent with no existing native session. For example, the question “Should existing JSON exports remain available?” can be followed by “No, continue.”, while the rebuilt prompt contains only the answer. The question remains in attention history; this report concerns its absence from the supplied continuation context.

The extraction itself appears deliberate. The existing [attention integration test](https://github.com/theam/facility/blob/c14bc6db34862e2e3135f9f6c0cf44370a998df8/services/api/test/turn-dispatcher.integration.test.ts#L416) expects the stripped conversation body and replies with the same builder. Separately, the [workspace end-to-end test](https://github.com/theam/facility/blob/c14bc6db34862e2e3135f9f6c0cf44370a998df8/services/api/test/facility-012.e2e.integration.test.ts#L416) supports switching to an architect/Claude agent with `nativeSessionId: undefined`. Neither test checks the question-plus-reply context on that combined transition.

## How to reproduce

The read-only script below executes the unchanged `agentOutcome`, `buildPrompt` and prompt helpers from this revision with three synthetic conversation messages. It models the persistence shape shown by the source and existing integration test. It does **not** execute a database, dispatcher integration or native agent.

Save it as `attention-question-repro.mjs`, then run against a checkout at the revision above:

```sh
node attention-question-repro.mjs /path/to/facility
```

Tested with Node 22.22.3; built-in modules only. Node emits an experimental warning for `stripTypeScriptTypes`. The source hash assertion prevents silently running this reproducer against a different implementation.

<details>
<summary>Runnable source-function reproduction</summary>

```js
import assert from 'node:assert/strict';
import { createHash } from 'node:crypto';
import { readFileSync } from 'node:fs';
import { stripTypeScriptTypes } from 'node:module';
import { resolve } from 'node:path';

// Read-only source-function reproduction. Tested with Node 22.22.3.
// Usage: node attention-question-repro.mjs /path/to/facility
// This models persisted messages; it does not execute a database or native agent.
const root = resolve(process.argv[2] ?? process.cwd());
const source = readFileSync(resolve(root, 'services/api/src/turns/dispatcher.ts'), 'utf8');
assert.equal(createHash('sha256').update(source).digest('hex'),
  '58e25034435686245325403163db8a83adb01fcf0512f16fa8ed43c2bedee93b',
  'Expected dispatcher.ts from c14bc6db34862e2e3135f9f6c0cf44370a998df8');

function extract(name) {
  const start = source.indexOf('\nfunction ' + name + '(');
  assert.ok(start >= 0, 'Missing source function: ' + name);
  const tail = source.slice(start + 1);
  const next = tail.search(/\nfunction \w+/);
  return next < 0 ? tail : tail.slice(0, next);
}
const { agentOutcome, buildPrompt } = new Function(
  stripTypeScriptTypes(['agentOutcome', 'buildPrompt', 'actorLabel', 'truncateStart']
    .map(extract).join('\n'), { mode: 'strip' }) + '\nreturn { agentOutcome, buildPrompt };',
)();

const question = 'Should existing JSON exports remain available?';
const answer = 'No, continue.';
const outcome = agentOutcome('I need one decision.\n' +
  '<facility-needs-attention>' + question + '</facility-needs-attention>');
const messages = [
  { seq: 1, role: 'user', turnId: 'first', body: 'Add CSV export.' },
  { seq: 2, role: 'agent', turnId: 'first', body: outcome.output },
  { seq: 3, role: 'user', turnId: 'fresh', body: answer },
].map(row => ({ ...row, actor: { type: 'system', id: 'fixture' } }));
const promptFor = rows => buildPrompt({ prompt: 'Continue the story.' },
  { title: 'CSV export', provider: 'ad_hoc', externalId: 'fixture' }, null, rows, 'fresh');
const prompt = promptFor(messages);
const control = promptFor(messages.map(row => row.seq === 2
  ? { ...row, body: outcome.output + '\n' + question } : row));

assert.equal(outcome.attention, question);
assert.equal(outcome.output, 'I need one decision.');
assert.equal(prompt.includes(answer), true);
assert.equal(prompt.includes(question), false);
assert.equal(control.includes(question), true);
assert.equal(control.includes(answer), true);
console.log(JSON.stringify({
  revision: 'c14bc6db34862e2e3135f9f6c0cf44370a998df8',
  attentionExtracted: outcome.attention,
  conversationOutput: outcome.output,
  rebuiltPromptHasAnswer: prompt.includes(answer),
  rebuiltPromptHasQuestion: prompt.includes(question),
  questionInConversationControl: control.includes(question),
  scope: 'Unchanged source helpers; modelled persisted messages; no database or native agent',
}, null, 2));
```

</details>

The assertions pass with:

```json
{
  "attentionExtracted": "Should existing JSON exports remain available?",
  "conversationOutput": "I need one decision.",
  "rebuiltPromptHasAnswer": true,
  "rebuiltPromptHasQuestion": false,
  "questionInConversationControl": true
}
```

The control puts the same question in the ordinary conversation message and confirms it reaches the prompt.

## Expected behavior

A fresh agent continuing after a blocking question should receive usable context connecting the question to the reply, without relying on the earlier agent's native history. The attention display, current-turn message boundaries and existing redaction should remain intact.

A focused regression could extend the existing fake-engine attention test: send a short reply to another valid agent manifest with no matching session, then check that the captured request has no prior native session and contains both question and answer. That combined integration case has not been executed in this report.

## Evidence and scope

- [Question extraction](https://github.com/theam/facility/blob/c14bc6db34862e2e3135f9f6c0cf44370a998df8/services/api/src/turns/dispatcher.ts#L675), [separate completion and attention writes](https://github.com/theam/facility/blob/c14bc6db34862e2e3135f9f6c0cf44370a998df8/services/api/src/turns/dispatcher.ts#L269), and [prompt construction](https://github.com/theam/facility/blob/c14bc6db34862e2e3135f9f6c0cf44370a998df8/services/api/src/turns/dispatcher.ts#L589).
- Same-agent native history, attention history, workspace notes or other retrieval may compensate. This does not show that every reply fails or that the question is deleted or unrecoverable.
- This is distinct from [#306](https://github.com/theam/facility/issues/306) / [#307](https://github.com/theam/facility/pull/307), which concern a queued request omitting a predecessor's later response. Here the question is removed from the ordinary message before prompt sequence filtering.

**Deployment:** Other — source-function reproduction, not a running Facility deployment.

**Version:** `c14bc6db34862e2e3135f9f6c0cf44370a998df8`.

haiderysfzi:
## What changes

A prompt rebuilt for an agent with no native session now carries the
`<facility-needs-attention>` question that the human's reply is answering. The
question is rendered against the agent message that asked it, so the transcript
reads question then answer.

## Why

Closes #344.

`agentOutcome` (`dispatcher.ts:669-674`) strips the question out of the output, and
`completeTurn` persists only the stripped body, so the question survives on the
attention item alone. `buildPrompt` builds its transcript purely from
`story_messages`. An engine resuming its own native session still holds the
question in context, which is why this is invisible on the common path. When the
reply goes to a different agent, the session lookup keys on
`agentName + engine + model` (`dispatcher.ts:169-177`), no session matches, and the
rebuilt prompt is all the agent gets: `No, continue.` with nothing it answers.

Two details are worth naming, because both are easy to get wrong.

Attention status is deliberately not filtered. Dispatch resolves the item as
`replied` at `dispatcher.ts:100-107`, *before* the prompt is built, so an
`status === "open"` filter would drop precisely the question being answered.

The question is placed on the agent message only. Both the user and the agent
message of a turn carry that turn's `turnId` (`stories/service.ts:367`), so keying
on `turnId` alone renders the question twice and once in the wrong place.

Only `agent_waiting` items are rendered; `queued_turn_dispatch_error` is an
operational failure rather than a question. Details are already redacted at write
time via `redactString`, so nothing new is exposed.

## Verification

New integration test: a builder asks a question, the reply is routed to a
different agent, and the rebuilt prompt is asserted to contain the question, the
answer, the question exactly once, and in that order. It also asserts
`nativeSessionId` is undefined, so the test fails rather than silently passing if
the fresh-session premise ever stops holding.

Mutating `dispatcher.ts` on this branch, running `turn-dispatcher.integration.test.ts`:

| mutation | new test | other 9 tests |
|---|---|---|
| drop the question rendering | fails | pass |
| filter attention to `status === "open"` | fails | pass |
| drop the `role === "agent"` guard (question renders twice) | fails | pass |

Each mutation is caught only by the new test, so the assertion is load-bearing
rather than incidental. The middle row is the one the existing suite could never
have caught.

Full runs, `services/api`, split across the databases the suite expects:

- `facility_test` (all but the e2e journey): 43 files passed, 407 tests.
- `facility_ws` (`facility-012.e2e.integration.test.ts`): 2/2, exit 0.
- `pnpm verify`: exit 0. Guards 2/2, audit clean against the ignore list.

One pre-existing failure, `test/oauth.test.ts` (`500` on dev login), is unrelated
and reproduces without this change. Same command, same freshly created database,
this branch stashed: `main` gives 406 passed with that same single failure, this
branch 407 with the same single failure. The delta is exactly the one added test.

## Note for @Yeeway69

The reproducer in the issue asserts a sha256 of `dispatcher.ts` to pin the
revision, and the file has changed since `c14bc6d` (in `b2a9b47`), so the script
now aborts on that assertion rather than running. The defect itself was untouched
by that commit and reproduces on current `main`. Thanks for isolating the seam so
precisely, it made this a small change.

### DIFF
--- a/services/api/src/turns/dispatcher.ts
+++ b/services/api/src/turns/dispatcher.ts
@@ -1,6 +1,7 @@
 import { type AgentManifest, AgentManifestSchema } from "@facility/agents";
 import { newId } from "@facility/core";
 import {
+  attentionItems,
   engineSessions,
   type FacilityDb,
   stories,
@@ -108,7 +109,7 @@ export class TurnDispatcher {
         });
       }
       await appendTurnEvent(this.db, { ...eventBase, type: "turn.started", data: {} });
-      const [story, workspace, conversation, messages] = await Promise.all([
+      const [story, workspace, conversation, messages, attention] = await Promise.all([
         this.scopedStory(input.orgId, input.projectId, turn.storyId),
         this.activeWorkspace(input.orgId, input.projectId, turn.storyId),
         this.db
@@ -134,6 +135,17 @@ export class TurnDispatcher {
             ),
           )
           .orderBy(asc(storyMessages.seq)),
+        this.db
+          .select()
+          .from(attentionItems)
+          .where(
+            and(
+              eq(attentionItems.orgId, input.orgId),
+              eq(attentionItems.projectId, input.projectId),
+              eq(attentionItems.storyId, turn.storyId),
+            ),
+          )
+          .orderBy(asc(attentionItems.createdAt)),
       ]);
       if (!workspace || !conversation)
         throw new Error("story workspace or conversation is missing");
@@ -202,7 +214,7 @@ export class TurnDispatcher {
         turnId: turn.id,
         manifest,
         workspace: workspaceLocator(workspace),
-        prompt: buildPrompt(manifest, story, conversation.summary, messages, turn.id),
+        prompt: buildPrompt(manifest, story, conversation.summary, messages, attention, turn.id),
         cwd: prepared.primaryCwd,
         nativeSessionId: session?.nativeSessionId,
         environment: prepared.processEnvironment,
@@ -593,6 +605,7 @@ function buildPrompt(
   story: typeof stories.$inferSelect,
   summary: string | null,
   messages: Array<typeof storyMessages.$inferSelect>,
+  attention: Array<typeof attentionItems.$inferSelect>,
   turnId: string,
 ) {
   const currentSequence = messages.find(
@@ -601,10 +614,16 @@ function buildPrompt(
   const relevant = messages.filter(
     (message) => currentSequence === undefined || message.seq <= currentSequence,
   );
+  const questions = askedQuestions(attention);
   const transcript = relevant
-    .map(
-      (message) => `${message.role.toUpperCase()} (${actorLabel(message.actor)}):\n${message.body}`,
-    )
+    .map((message) => {
+      const body = `${message.role.toUpperCase()} (${actorLabel(message.actor)}):\n${message.body}`;
+      // Both the user and the agent message of a turn carry its `turnId`, so the
+      // question is placed on the agent message that actually asked it.
+      const asked =
+        message.role === "agent" && message.turnId ? questions.get(message.turnId) : undefined;
+      return asked ? `${body}\n\n${QUESTION_HEADING}\n${asked.join("\n")}` : body;
+    })
     .join("\n\n");
   return [
     manifest.prompt,
@@ -616,6 +635,27 @@ function buildPrompt(
   ]
     .filter(Boolean)
     .join("\n\n");
+}
+
+const QUESTION_HEADING = "Question this turn asked the human:";
+
+// `agentOutcome` strips the <facility-needs-attention> question out of the agent's
+// message before it is persisted, so the question survives only on the attention
+// item. An engine resuming its own native session still holds it in context, but a
+// prompt rebuilt for a session that does not exist — a reply routed to a different
+// agent, or a replaced session — would otherwise show the human's answer with
+// nothing it answers. Status is deliberately not filtered: dispatch resolves the
+// item as `replied` before the prompt is built, so an open-only filter would drop
+// exactly the question being answered.
+function askedQuestions(attention: Array<typeof attentionItems.$inferSelect>) {
+  const questions = new Map<string, string[]>();
+  for (const item of attention) {
+    if (item.kind !== "agent_waiting" || !item.turnId || !item.detail) continue;
+    const asked = questions.get(item.turnId);
+    if (asked) asked.push(item.detail);
+    else questions.set(item.turnId, [item.detail]);
+  }
+  return questions;
 }
 
 function actorLabel(actor: unknown) {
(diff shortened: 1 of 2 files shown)


adrian-lorenzo (approved):
Thanks for fixing this.

The change restores the missing context in #344, and the regression test covers the handoff well. I verified the fix and found no blocking issues.

haiderysfzi:
> Thanks for fixing this.
> 
> The change restores the missing context in #344, and the regression test covers the handoff well. I verified the fix and found no blocking issues.

Thanks for the review! Let me know if you'd like any changes before merge, or if there's a next issue to grab.

Issue #376, opened 2026-09-12
Automate stable workspace public-site provisioning


alvaroloes:
## Goal

Automatically provision a dedicated stable public site per workspace/service.
Today FACILITY_PREVIEW_SITES supports exact bindings, but an operator must create
the distribution, update configuration and refresh the API. Merging Compose code
alone does not create a public origin.

This remains a **follow-up after the manual-site integration pilot** for AIFindr D3
(theam/aifindr-dev-environment#30). Generic story facts, integrationState and GitHub
notifications are now in #374 / #375; the project-owned Auth0 consumer is in
theam/aifindr-dev-environment#31. They are not part of provider provisioning here.

## Incremental plan

1. Add a provider-neutral stable-site provision/reconcile contract, initially for
   the CloudFront path supported by the internal deployment. Bind exact
   org/project/workspace/service idempotently and persist provider identity.
2. Create a disabled distribution/per-site credential, register and read back the
   mapping, refresh configuration, then publish the configured origin after gateway
   routing/authentication checks. Do not require the app's Auth0 login before
   reporting the origin: its consumer needs that origin to register the callback.
3. Keep site/storage separate from compute. Suspension retains identity. Never
   reassign browser origins across workspaces; coordinate retirement so external
   cleanup finishes before removing its configuration/tombstone.
4. Make site changes observable by the generic lifecycle notification mechanism.
   No Auth0-specific script or provider credentials in agent workspaces. Consumers
   own business policy and use Facility's generic integration state if needed.
5. Add a scoped operator workflow in the private deployment repository. Existing
   release-on-merge remains; its role currently does not provision CloudFront sites
   or write preview configuration. Review exact IAM, cost and secret targets with
   deterministic fake-provider/denial/recovery tests before live application.

## Acceptance

A fresh story obtains an isolated root-origin site without manual per-story cloud
work. Sleep/resume retains identity. Public origin publication enables project
integrations to complete login setup; retired stories can clean owned registrations
without destroying retained work. A second story is isolated. Failed/partial
provisioning and retries cannot duplicate, cross-bind or expose a workspace.

No automatic provider or real D3 preview is claimed by creating this issue.

alvaroloes:
Approved next slice: #386 implements opt-in Facility login on the native Vercel Sandbox URL, replacing the CloudFront-first manual-site prerequisite for the AIFindr pilot. It reuses existing lifecycle site facts and workspace.updated dispatch; project-owned callback consumers remain separate.

[...]

PR #375, opened 2026-09-12
feat(stories): add integration state and lifecycle notifications

alvaroloes:
## Summary

Closes #374. Revised after the AIFindr integration design review.

- Extend the **existing story GET** with generic `lifecycle` facts, GitHub
  freshness, workspace identity and scoped configured public sites. Remove the
  registration-specific GET previously proposed in this PR.
- Store bounded, namespaced `story.integrationState` JSON in Facility, with a
  revision-checked PATCH (`stories:write`), 16 KiB limit and normal tenant/project
  isolation. Retain it across suspension, archive and soft deletion. No secrets.
- Publish lightweight `facility.story.updated` / `facility.workspace.updated`
  repository-dispatch events to the project's primary repository. A durable worker
  cursor retains pending event identity, leases, retries/backoff and sanitized errors.
  Consumers fetch current state; HTTP acceptance does not prove a workflow ran.
- Metadata writes do not change story timestamps or emit lifecycle events, avoiding
  consumer loops. Initial/coarse observation runs once per minute in fair batches;
  intermediate transitions may be coalesced. This is not a full audit event stream.
- Document the generic contract in the API reference and `docs/story-integrations.md`.
- Address the PR-backed story review: expose source identity and PR mirror freshness;
  use the exact source PR, never a same-branch or child-repository substitute.

## Boundaries and adoption

One additive v0.12 migration (`0007_story_integrations.sql`) is required. Existing
API fields remain; the removed registration-specific endpoint was only in this
unmerged PR. GitHub App Contents: write is required for repository dispatch.
API and worker must share `FACILITY_PREVIEW_SITES`; refresh both on manual binding
changes. Apply the migration before deploying the new API/worker.

Facility owns generic storage/events, not Auth0 policy. No Auth0 SDK/key, cloud
provisioning, runtime wake, agent change, public-site automation or deployment here.
An empty site list/stale issue/404 is not a cleanup instruction. Associated PR merge
can mark a story done; this is not downstream production-deployment evidence.
State CAS is not a lock on shared third-party resources; consumers serialize writers.
No namespace-specific ACL is promised beyond scoped `stories:write`.

AIFindr consumer: https://github.com/theam/aifindr-dev-environment/pull/31.
Automatic public-site provisioning remains #376 and is intentionally separate.

## Verification

- **53 targeted tests pass** across lifecycle/API/database, backlog/routes and phase:
  PR-backed stories after a completed turn, open/draft/closed/merged/reopen, sync-time
[...]

PR #386, opened 2026-09-15
feat(previews): add Facility login for native Vercel workspace URLs

alvaroloes:
## Summary

Refs #376. Add opt-in Facility login on the native HTTPS URL already provided by
Vercel Sandbox, without provisioning a per-story CloudFront distribution.

- The runner gateway redirects visitors to Facility login, exchanges a 60-second
  browser-bound one-use code server-to-server, and sets a different host-only
  preview session cookie. Facility's main login cookie never reaches the app.
- Recheck current membership/role, expiry/revocation and exact workspace/service/
  origin on requests and WebSocket handshakes. The callback code cannot become a
  reusable cookie or enter the legacy preview exchange.
- `previews:read` allows opening prepared/running native previews without waking
  compute, issuing repository credentials or running setup/start/agents. Existing
  `workspaces:execute` users retain prepare/wake behavior. No per-story ACL.
- Publish provider-verified origins through existing lifecycle site facts and
  `workspace.updated` notifications, reusing project-owned integration consumers.
- Keep Docker and legacy proxy behavior behind default-off
  `FACILITY_NATIVE_PREVIEWS`. Based on latest main `b8e26ea`, preserving #323's
  execution-permission recheck for all legacy proxy sessions.

## Deployment and limits

Apply migration `0008_native_preview_sessions.sql`, publish the matching runner
image, deploy API/worker, then explicitly opt in on both. Old persistent images
fail capability checks; changing the image setting does not upgrade existing disks.
No production activation, workspace recreation, Auth0 write or agent run is part
of this PR. Start validation with a disposable story, preserving retained work.

Human permissions use Facility's existing organization-role model; this does not
invent project-level memberships. The cookie is workspace-bound, not a new
per-preview permission. "Read" means access to the interactive app, not limiting
the app itself to GET requests.

The in-VM gateway protects visitor access, not against a hostile agent controlling
that same VM. An independent edge is needed for that stronger boundary. Existing
API rate limits apply to authorization requests.

Live native URL stability across suspend/resume is **not yet verified**. Current
provider origins are published as facts; changes invalidate old sessions and must
be reconciled by external callback consumers. Keep #376 open for live acceptance
and any stable-alias follow-up. Sleeping direct URLs do not wake compute.

Operator guide and rollback: `docs/native-workspace-previews.md`.

## Contract generation

Regenerating OpenAPI/SDK also caught an existing recursive `z.json()` reference
that becomes invalid when embedded in Swagger. Preserve its runtime JSON
validation behind a refinement while emitting the valid arbitrary-JSON schema;
add a regression for both JSON rejection and reference-free embedding. Generated
files include the previously ungenerated integration-state contract from #375.
The initial minimum-node CI run caught the corresponding missing SDK route-manifest
entry. That entry is now included; all 10 SDK tests and the complete non-database
test command pass locally. The updated commit requires a fresh CI result.

## Validation

- 133 API tests passed across 11 focused suites: native browser flow, real
  PostgreSQL session persistence/atomic replay denial, local HTTP/WebSocket
  forwarding, SDK/bootstrap capability, legacy previews, environment preservation,
  configuration and story integrations.
- 8 web tests passed across preview launch, workspace controls and story page.
- Expanded `pnpm test:uncached` passed: all 119 script tests and all non-database
  package suites, including 90 web, 10 SDK, 14 MCP, 11 docs, 10 core and 10 agent tests.
- Full monorepo typecheck, lint, repository guards, unused-reference check and
  migration compatibility check passed.
- OpenAPI and SDK regenerated successfully using disposable local fixtures.
- Real Vercel login, application OAuth, two-story isolation and sleep/resume
  persistence remain post-deployment checks. No live credentials are needed by
  default tests. Full repository CI remains authoritative for merge.


services/api/src/routes/v1/native-workspace-previews.ts
```diff
@@ -0,0 +1,89 @@
...
+      const { workspaceId, service } = request.params as z.infer<typeof Params>;
+      const { challenge } = request.query as { challenge: string };
+      reply.header("cache-control", "no-store").header("referrer-policy", "no-referrer");
+      if (!request.principal) {
+        const login = new URL("/api/auth/login", config.webUrl ?? config.publicUrl);
+        login.searchParams.set(
+          "returnTo",
+          `/api/workspace-preview-login/${workspaceId}/${service}?challenge=${encodeURIComponent(challenge)}`,
+        );
+        return reply.redirect(login.toString());
```
alvaroloes, line 41: [P1] Allow the preview return path through Facility's login handler

For a visitor without an active Facility session, this `returnTo` never survives `/auth/login`: `safeReturnTo()` in `services/api/src/routes/auth.ts` only accepts `/oauth/interaction/<id>` and replaces every other path with `/`. Reproducing the two redirects through `buildApp()` shows `facility_oauth_state` contains `returnTo: "/"`. After successful identity-provider login, the browser therefore lands on Facility's home page instead of returning here to obtain the preview grant. Add a narrowly validated native-preview return path to that allowlist and an integration test that continues through the real login handler; the current test stops after asserting the initial redirect.

services/api/src/routes/v1/native-workspace-previews.ts
```diff
@@ -0,0 +1,89 @@
...
+          ...(request.body as { sessionId: string; code: string; verifier: string }),
+        },
+        request.headers["x-facility-preview-token"] ?? "",
+      );
+    },
+  );
+  app.post(
+    "/workspace-preview-native/:workspaceId/:service/authorize",
+    {
+      config: { public: true },
```
alvaroloes: [P2] Give per-resource authorization an appropriate rate-limit budget

This route inherits `buildApp()`'s production limit of 200 requests/minute per IP, while the gateway calls it for every asset, application request and WebSocket handshake. The gateway does not forward visitor IPs, so visitors using the same gateway/egress share that budget. A development page with many modules, repeated reloads or concurrent testers can exhaust it during normal use. With the real production limiter and an authorization stub accepting every credential, 205 gateway resource requests produce 200 successes followed by five 503s (the API's 429 is converted by the gateway). Configure a suitable authenticated authorization limit while retaining per-request revocation checks, and test this route with the production limiter enabled.

runner/facility-preview-gateway.mjs
```diff
@@ -1,70 +1,265 @@
...
+          3600,
+          Math.floor((new Date(result.expiresAt).getTime() - Date.now()) / 1000),
+        );
+        if (!Number.isFinite(age) || age <= 0) throw new Error("Expired session");
+        response
+          .writeHead(302, {
+            location: "/",
+            "set-cookie": [
+              setCookie(COOKIE, `${sessionId}.${result.accessToken}`, age),
+              setCookie(NONCE, "", 0),
```
alvaroloes, line 217: [P2] Restore the original application URL after preview sign-in

The gateway does not retain the URL that triggered login, and this callback always redirects to `/`. Opening a shared deep link without a preview cookie therefore loses its path/query even when the user is already signed into Facility. It also drops an application's OAuth callback if the preview session expires while the browser is away at the application identity provider: `/auth/callback?code=...&state=...` becomes `/`, so the application never receives its code/state. I reproduced that redirect with the real gateway and a successful exchange. Preserve a validated same-origin relative destination in the browser-bound login state and restore it here.

alvaroloes (commented):
Reviewed all 36 changed files at `531faa3844b418ac4407f9f1ceb2f0bb93c46b03`, including browser login, persisted session authorization, tenant/service/origin binding, HTTP/WebSocket forwarding, Vercel bootstrap, lifecycle publication, migration, UI and generated contracts. Three reproducible integration issues are detailed inline and should be fixed before merging.

Validation: 135 API tests across 11 focused suites, 8 web tests and 10 SDK tests passed locally; migration compatibility and diff checks passed. Three additional isolated regression tests fail on this head: the real login transaction discards the preview return destination; gateway reauthentication discards the application path/query; and the production API limiter causes 5 of 205 valid preview resource requests to return 503. These tests use local servers/test doubles and disposable PostgreSQL, with no live provider credentials.

Separately, a local merge check against current `main` (`9c186c9`) reports a conflict in `services/api/src/workspaces/vercel.ts`. Real Vercel/browser OAuth and suspend/resume acceptance remain unverified.

services/api/src/routes/v1/native-workspace-previews.ts
alvaroloes: Fixed in 2723ef9. Facility login now allows only the exact native-preview continuation (validated workspace/service plus one challenge parameter), while preserving the existing OAuth-interaction allowlist and rejecting unrelated/absolute destinations. The native integration fixture now uses buildApp(), real login/callback/session middleware and a deterministic fake GitHub identity provider. It follows the anonymous preview request through sign-in and back to the grant exchange, rather than stopping at the first redirect. Added pure allowlist denial tests as well.

services/api/src/routes/v1/native-workspace-previews.ts
alvaroloes: Fixed in 2723ef9 with a dedicated, bounded /authorize rate-limit bucket: 6,000 requests/minute/IP, independent of the ordinary production 200/minute bucket. No caller-controlled forwarded-IP/session key, no permission caching, and no global limiter removal. The initial ceiling is intentionally still shared by gateway egress and documented for pilot tuning. With buildApp() and the real gateway/DB, 205 resource requests succeed, immediate role removal denies the next request, request 6,001 hits the dedicated limit, and ordinary API requests still hit their unchanged 200 limit. Spoofed X-Forwarded-For does not reset the budget.

runner/facility-preview-gateway.mjs
alvaroloes: Fixed in 2723ef9. The gateway retains the original same-origin relative path/query in its host-only HttpOnly login cookie, with an HMAC binding it to the verifier, gateway secret and a two-minute expiry. Application callback data stays on the preview host and is not included in Facility/GitHub OAuth redirects. Successful exchange restores that path; tampered/expired state fails closed, and unsafe/reserved/oversized destinations fall back to /. Unit tests cover redirect validation, tampering, expiry and wrong gateway. Real-gateway integration follows a deep link and an application OAuth callback through preview reauthentication, preserves app cookies/query parameters, and rejects replay.

alvaroloes (commented):
Re-reviewed `2723ef956aa211394335ae9b05f496b4d55a74b4`. No new actionable findings; all three issues from my previous review are verified fixed.

- The real Facility login transaction preserves the narrowly validated native-preview continuation, and the integration suite follows login/callback/session middleware through to the preview grant.
- The dedicated authorization budget handles the 205-resource reproduction while preserving immediate permission revocation, bounded denial traffic and the ordinary API limit.
- Signed browser login state restores application paths/query parameters, including OAuth callbacks, and rejects tampered, expired and unsafe destinations.

Validation on this head: 174 API tests across 15 suites (including my three independent regressions from the previous review), 8 web tests and 10 SDK tests passed. Dependency builds, migration compatibility and diff checks passed. I also checked the merge resolution: both Vercel imports are retained and the branch merges cleanly with `main` at `8f7f6ef`.

The code review is favorable for the documented opt-in scope. At the latest CI check, `verify`, `minimum-node` and title/commit checks passed; `self-host-build` and `workspace-e2e` were still running. Real Vercel/browser OAuth and suspend/resume acceptance remain pending as documented. Resolving the three previous review threads.

PR #397, opened 2026-09-16
feat(workspaces): supply native preview origins before setup

alvaroloes:
## Why

Some projects need the workspace's native HTTPS origin before building their application (OAuth redirect URLs, cookies, frontend runtime config). Vercel already knows that origin when it creates/wakes the sandbox, but Facility previously made it available to integrations only after application readiness. This lets consuming projects configure origin-dependent settings in a single setup phase.

## Change

- Add optional runtime `previewOrigins` metadata and the opt-in, runtime-owned `FACILITY_PREVIEW_ORIGINS` environment variable (JSON service-name to origin map).
- Vercel reuses the SDK URL and existing credentialed gateway binding check before project setup. It does not start a test app or publish readiness.
- Inject the map into declared project hooks, including resume. Reject managed/operator/repository-credential overrides of this reserved value.
- Preserve normal start/readiness, final gateway verification, endpoint persistence and existing lifecycle notifications. Reject origin drift before publication.
- No new endpoint, lifecycle event, acknowledgment, schema migration or runner change. No project is automatically opted in. Manifests that do not request the variable retain their current startup path.

External integrations, such as authorization configuration, remain project-owned and asynchronous after `facility.workspace.updated`. Application readiness does not imply that external integration setup has completed; this contract does not introduce an additional startup phase or completion acknowledgment.

## Validation

- 36 focused tests passed: Vercel unit coverage plus project-environment integration tests on disposable local PostgreSQL (zero skips).
- Tests cover origin availability before setup, no endpoint publication on readiness failure, resume, origin drift, project opt-out, invalid gateway binding, unsupported runtimes, reserved-value overrides and unchanged legacy startup.
- API TypeScript check, scoped Biome check and `git diff --check` passed.
- External providers are deterministic fakes; no live provider or project mutation or deployment was performed.

## Rollout

Merge/deploy this API/worker change before activating any consuming project manifest that declares `FACILITY_PREVIEW_ORIGINS`. This PR does not claim remote application acceptance or stable URLs across real suspend/resume.

Related: #376. Do not close that runtime-acceptance issue from source tests alone.


Lob26:
Question on the origin check in `startServices`: on `startPrepared`, `withDeclaredEnvironment` looks up the origins immediately before `expose`, so both reads are current and the check can only catch drift within a single call. A resume where the sandbox comes back on a new origin — setup skipped because `setupChecksum` hashes only the manifest and HEAD — passes it and publishes.

The PR body scopes stable URLs across suspend/resume out, which is fair, but the docs line "If the final native binding differs from the origins supplied to setup, Facility rejects publication" reads as covering that case. Worth narrowing it to drift during a single prepare?

alvaroloes (commented):
All good

alvaroloes:
@Lob26

> Worth narrowing it to drift during a single prepare?

Good catch — you're right. This check compares the origins resolved for the current `prepare` / `startPrepared` invocation with the final `expose` result, not with the origins used by the original setup.

I'll narrow the wording to make that boundary explicit and clarify that the map is resolved again on resume. This PR does not guarantee origin stability across suspend/resume or automatically invalidate setup when an origin changes. Projects that persist origin-dependent configuration need to validate it in their resume hooks.

PR #404, opened 2026-09-17
feat(workspaces): support per-project CPU and memory limits

alvaroloes:
## Summary

Allow projects to declare optional `environment.resources` in `.facility.yml`, so multi-service development stacks can request enough CPU and memory without changing the shared default.

```yaml
environment:
  resources:
    cpu: 4
    memory_mb: 8192
  start: docker compose up -d
```

## Behavior and compatibility

- One mapping is shared by API/UI/MCP starts, GitHub triggers and scheduled stories.
- Existing manifests still use 2 vCPU / 4096 MiB. No migration, deployment setting or global default changes.
- Size is captured when a **new story workspace** is created. Existing workspaces retain their allocation; retries, resume and Clean setup do not resize or discard data.
- Strict bounds reject malformed/incomplete resource requests. Vercel's pure creation preflight validates its fixed 2048 MiB-per-vCPU ratio before new workspace settings are committed, and is reused at allocation. A rejected pair can be corrected and retried for the same story without pinning invalid configuration. Existing workspaces are not revalidated against changed manifest settings. Provider/account limits still apply; Docker applies CPU and memory separately.
- Larger allocations may cost more and remain a repository-reviewed choice.

## Verification

- 110 tests passed across resource parsing/mapping, Vercel runtime, API story creation, GitHub/schedule automation, persistent story lifecycle and turn dispatch. Integration suites use a disposable local PostgreSQL instance and fake providers; no live credentials.
- Regression covers omitted defaults, invalid requests, all creation paths and a manifest size change replaying against an existing story without resizing it. New GitHub/schedule recovery tests failed before the service preflight change and pass after it: invalid settings roll back without allocation, the corrected pair starts the same identity, and subsequent resume retains its original allocation.
- API typecheck and targeted Biome checks passed; `git diff --check` clean.
- Corrected the route test fixture to explicitly opt into the demo organization/user it requires on a fresh database.

## Rollout

Deploy Facility before downstream repositories add this field: older versions deliberately reject unknown manifest fields. This PR alone does not opt any project into a larger machine.


services/api/src/workspaces/vercel.ts
```diff
@@ -52,6 +52,18 @@ export class VercelWorkspaceRuntime implements WorkspaceRuntime {
...
+    const cpu = input.resources?.cpu ?? 2;
+    if (
+      !Number.isInteger(cpu) ||
+      cpu < 1 ||
+      cpu > 32 ||
+      (input.resources && input.resources.memoryMb !== cpu * 2_048)
+    ) {
+      throw new WorkspaceRuntimeError(
+        "workspace_resources_invalid",
+        "Vercel requires 1–32 integer vCPUs with exactly 2048 MiB per vCPU; account limits also apply",
```
alvaroloes, line 63: [P2] Validate provider constraints before persisting the workspace

This check runs after `StoryWorkspaceService.start()` has committed the workspace configuration. With `{cpu: 4, memory_mb: 4096}`, parsing succeeds but creation fails here without allocating a sandbox. Correcting the manifest to `memory_mb: 8192` and retrying the same issue or scheduled story still fails: `start()` reuses the error-state row and passes its original 4096 MiB back into this check. I reproduced both attempts using the invalid pair, while a fresh story with the corrected pair succeeds. Validate the provider-specific constraints before committing a new workspace, or allow an explicitly rejected, never-allocated workspace to adopt the corrected configuration on retry; existing allocated workspaces should keep their size.

alvaroloes (commented):
Reviewed `fd75aae03575ae02511cd15915b86b0a3820a4f7`. One P2 recovery issue is documented inline: a resource pair rejected before provider allocation remains pinned to the story, so correcting the manifest does not unblock that story.

Validation: 93 existing tests passed across manifest resources, Vercel, API creation, GitHub/schedule automation and persistent workspace lifecycle. A reviewer-added integration test using disposable PostgreSQL and a fake provider reproduces the failed recovery; the corrected pair succeeds for a fresh story. Dependency builds, API typecheck, targeted Biome and `git diff --check` passed. CI is green. No live provider allocation or deployment was performed.

services/api/src/workspaces/vercel.ts
alvaroloes: Fixed in 9898f7d. Creation now runs the provider’s pure preflight inside the transaction, before a new workspace configuration is persisted; Vercel reuses the same validation at allocation. Invalid pairs roll back without allocating or pinning a workspace, so the corrected manifest can start the same issue/scheduled identity. Existing workspaces skip this new-config preflight and keep their recorded size on resume. No repair or resizing mechanism was added.

Added GitHub and schedule regression cases that failed before the service change and now pass, plus a no-provider-I/O preflight unit test. All 110 targeted tests pass with disposable local PostgreSQL and fake allocation; API typecheck, Biome and diff checks pass. Also merged current main into the branch without rewriting history.

